PR #190, opened 2014-12-05
Modify A::D::Query to use POE events rather than while(1)

zachthompson:
**Which issues (if any) does this change solve? (please link them here)**
n/a

**Briefly describe your fix. Are there any important details we should know about?**

* Uses POE's events correctly to generate the "infinite loop" effect
* Some other small changes for efficiency

**Are you having any problems? Do you need our help with anything?**
n/a

### DIFF
--- a/lib/App/DuckPAN/Query.pm
+++ b/lib/App/DuckPAN/Query.pm
@@ -6,6 +6,7 @@ use Data::Printer;
 use POE qw( Wheel::ReadLine );
 use Try::Tiny;
 
+# Entry into the module.
 sub run {
     my ( $self, $app, $blocks ) = @_;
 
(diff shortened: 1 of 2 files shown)


nilnilnil:
This looks great.

lib/App/DuckPAN/Query.pm
```diff
@@ -2,95 +2,125 @@ package App::DuckPAN::Query;
 # ABSTRACT: Main application/loop for duckpan query
 
 use Moo;
-
-my $query;
-my $history_path;
-
+use DDG;
+use DDG::Request;
```
moollaza: It seems that moving this here has cause all the Travis tests to fail because DDG::Request cannot be found in @INC. The problem with DDG is that it's not a CPAN Repository, instead its hosted on our own duckpan.org.

I'm not entirely sure how/why the `require...import` method makes a difference with regards to making DDG::Request accessible...

moollaza: Ah I get it. `App::DuckPAN::DDG` gets loaded by `App::DuckPAN` at https://github.com/duckduckgo/p5-app-duckpan/blob/f6dbd313e7efec927d109bd65b1f8b30c12dd6db/lib/App/DuckPAN.pm#L262

It's `require`d and `import`ed only once we need it to run Query.pm

@zachthompson you're much more experienced with Perl than me -- do you have any ideas about how to fix this somewhat convoluted situation? Should we just revert back to using `require` and `import` inside `run`?

moollaza: Sorry, that's wrong. I should have just gone to sleep last night ;)

The way DuckPAN ensures DDG is installed, is that it checks for it to be installed thanks to:
(check_requirements)[https://github.com/duckduckgo/p5-app-duckpan/blob/076547d8a020afcbbee880d76d16411a456d082e/lib/App/DuckPAN.pm#L375] which calls (check_ddg)[https://github.com/duckduckgo/p5-app-duckpan/blob/076547d8a020afcbbee880d76d16411a456d082e/lib/App/DuckPAN.pm#L490] and then prompts the user to run `duckpan DDG` if DDG is not installed.

moollaza: This is why Travis was able to build DuckPAN before and the tests all happily passed. This means then that our tests don't actually execute anything that requires DDG. Sigh.

We're going to need to fix that sooner or later...

zachthompson: @moollaza Yeah, I noticed that was causing travis issues after the fact.  We can change it back.  It's not too big of a deal since we're pretty sure these modules will be there.

moollaza: @zachthompson let's do that for now so tests pass as expected. Then we can work towards a better approach after

PR #203, opened 2015-01-23
Make sure DDG installed when testing DuckPAN

moollaza:
It seems that Travis tests are failing because DDG::Request cannot be found in @INC. I believe this is because Travis doesn't have DDG installed, so this should solve our problems.

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -105,3 +105,4 @@ debug = 0
 perl_version = 5.14
 perl_version = 5.16
 perl_version = 5.18
+after_install = duckpan DDG


moollaza:
I shouldn't write code after midnight. This fix obviously won't work. We can't use duckpan to pre-install DDG when building DuckPAN on a machine. I understand now that `App::DuckPAN::DDG` serves this exact purpose. It loads `DDG` before running any DuckPAN commands that need it and then `DDG` is `require`'d and `import`'ed as it's needed.

This functionality was mistakenly removed in #190. We likely need to bring it back or find a better solution.

moollaza:
Closing this for now. Going to think about how we should move forward

PR #205, opened 2015-01-23
App::DuckPAN::Query: revert to require DDG::* for travis

zachthompson:
Revert to 'require DDG::*' instead of 'use' so travis builds will hopefully succeed.


moollaza:
LGTM :+1: 

Thanks a lot @zachthompson

PR #209, opened 2015-01-26
DuckPAN Dev Versioning

moollaza:
When DuckPAN is built and installed via `Dist::Zilla`, the `PkgVersion` plugin injects a line that sets the package's `$VERSION` based on the current Git Tag for the DuckPAN repo. Previously, DuckPAN would set `$VERSION` to `'9.999'` if it wasn't set by PkgVersion, which only occurred when DuckPAN was run 'in place' from the repo, i.e. `./bin/duckpan -I /lib`.

A bug was recently fixed in `PkgVersion` which forces it to **not** inject a `$VERSION = ` line for any files that explicitly set a value for `$VERSION`. (We've just been getting lucky for the past few years).

This means the `our $VERSION ||= '9.999'` line in `App::DuckPAN` had to be removed and a "dev" version must be set in another manner.

This PR sets the `$VERSION` to `'9.999'` in those cases.

---

**Note:** The DDG (duckduckgo/duckduckgo) Module has the name `our $VERSION ||= '9.999'` line, which will soon be removed. Currently DuckPAN is the only software which expects DDG to ever have a version of `9.999`

/cc @russellholt @jagtalon @zachthompson @MrChrisW @jbarrett

### DIFF
--- a/lib/App/DuckPAN.pm
+++ b/lib/App/DuckPAN.pm
@@ -100,7 +100,7 @@ option duckpan => (
 sub _ua_string {
   my ($self) = @_;
   my $class   = ref $self || $self;
-  my $version = $class->VERSION;
+  my $version = $class->VERSION; || 'dev';
   return "$class/$version";
 }
 
--- a/lib/App/DuckPAN.pm
+++ b/lib/App/DuckPAN.pm
@@ -100,7 +100,7 @@ option duckpan => (
 sub _ua_string {
   my ($self) = @_;
   my $class   = ref $self || $self;
-  my $version = $class->VERSION; || 'dev';
+  my $version = $class->VERSION || '9.999';
   return "$class/$version";
 }
 
--- a/lib/App/DuckPAN/Perl.pm
+++ b/lib/App/DuckPAN/Perl.pm
@@ -119,10 +119,11 @@ sub duckpan_install {
 				}
 			}
 		} elsif ($localver == $duckpan_module_version) {
-			$message = "You already have latest ($localver) version of $package";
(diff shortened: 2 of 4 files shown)


zachthompson:
@moollaza Need the 9.999 trick in check_ddg/check_app_duckpan in App::DuckPAN

moollaza:
@zachthompson I believe it's already there? https://github.com/duckduckgo/p5-app-duckpan/blob/master/lib/App/DuckPAN.pm#L472 -- or do you mean something else?

zachthompson:
@moollaza Lines 472/493 only check $installed_version but don't set it.  get_local_*_version will return undef, at least in my testing it did.  So, either $installed_version needs to be set or the return statement needs to be modified to "return $ok unless $installed_version" which amounts to the same thing.

The get_local_*_version subs seem like needless overhead, as they just delegate to get_local_version with the module name.  But we know the module we're after in the check_* subs and could just call it directly.

In setting the undefined versions to 9.999, are we overlooking the use case where the module isn't even installed?  Will it have no way to get installed?

moollaza:
>get_local_*_version will return undef, at least in my testing it did.

Not sure how that happened, it's returning '9.999' for me, as expected in my testing.

>In setting the undefined versions to 9.999, are we overlooking the use case where the module isn't even installed? Will it have no way to get installed?

I hadn't thought about that! I verified that it works correctly when DDG is not installed. The `eval` in `get_local_version` fails (the module isn't able to be loaded and `require`'d) and so the function `return`s before my check for `$v` and `$module`

I'll add a comment there to explain this behaviour.

Issue #126, opened 2014-10-02
DuckPAN Server: Tabs are overlapping each other


jagtalon:
![screen shot 2014-10-02 at 11 27 52 am](https://cloud.githubusercontent.com/assets/81969/4493346/bdba286e-4a48-11e4-8abb-c02814865a8c.png)

PR #212, opened 2015-01-27
Support built-in templates for Goodies

zachthompson:
Adds support for using the built-in Spice templates with Goodies.   Goodies can return a hash ref, e.g. [Goodie with built-in templating](https://github.com/zachthompson/zeroclickinfo-goodies/blob/goodie-template-v2/lib/DDG/Goodie/GoodieTemplateV2.pm), that essentially mimics the Spice.add call.

### DIFF
--- a/lib/App/DuckPAN/Web.pm
+++ b/lib/App/DuckPAN/Web.pm
@@ -325,14 +325,13 @@ sub request {
 				my $zci_wrapper = $root->look_down(id => "zero_click_wrapper");
 				$zci_wrapper->insert_element($zci_container);
 
-
 				my $duckbar_home = $root->look_down(id => "duckbar_home");
 				$duckbar_home->delete_content();
 				$duckbar_home->attr(class => "zcm__menu");
 				$duckbar_home->push_content(
 					HTML::TreeBuilder->new_from_content(
 						q(<li class="zcm__item">
-					    	<a data-zci-link="answer" class="zcm__link  zcm__link--answer is-active" href="javascript:;">Answer</a>
+							<a data-zci-link="answer" class="zcm__link  zcm__link--answer is-active" href="javascript:;">Answer</a>
 						</li>)
 					)->guts
 				);
@@ -345,7 +344,6 @@ sub request {
 				
 				# Make sure we only show one Goodie (this will change down the road)
 				last;
-
 			}
 			if ($result_type eq 'other') {
 				# Not Spice or Goodie, inject raw Dumper() output from into page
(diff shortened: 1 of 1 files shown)


lib/App/DuckPAN/Web.pm
```diff
@@ -285,11 +284,21 @@ sub request {
...
 					# There is no error-checking or support for non-auto-templates here.
 					my $structured = $result->structured_answer;
-					my $template_name = 'goodie_'.scalar @{$structured->{input}}.'_inputs';
-					my $json_string = encode_json({Answer => $structured});
-					$zci_container->push_content(HTML::TreeBuilder->new_from_content("<script>\$(window).load(function(){"
-								. "document.getElementById('zci-answer').innerHTML = DDG.exec_template('$template_name', $json_string);"
-								. "});</script>")->guts);
+					if(exists $structured->{templates}){ # user-specified templates
+						$page = $root->as_HTML;
+						$page =~ s|####DUCKDUCKHACK-CALL-NRJ####|;DDG.duckbar.future_signal_tab({signal:'high',from:'$structured->{id}'});|;
```
moollaza: We actually do this below starting at line 359. If you just create the script tag and it into the correct arrat, e.g. `@calls_nrj` this should work the same and then you don't have to both modifying the page here I believe?

moollaza: @zachthompson do you know if we plan to allow multiple Goodies now that they can have different tab names? Pushing into the `@calls_` arrays will also allow for that, if we want.

zachthompson: @moollaza I thought about taking that route but the needs of Goodies are a bit simpler and put together differently.  We would need to detect the difference and modify the construction below.  To me it seems like more work than is necessary.

We should try multiple goodies at some point but let's do it as a separate PR.

moollaza:
Tested and works perfectly. LGTM :+1:

lib/App/DuckPAN/Web.pm
```diff
@@ -285,11 +285,17 @@ sub request {
...
-					my $template_name = 'goodie_'.scalar @{$structured->{input}}.'_inputs';
-					my $json_string = encode_json({Answer => $structured});
-					$zci_container->push_content(HTML::TreeBuilder->new_from_content("<script>\$(window).load(function(){"
-								. "document.getElementById('zci-answer').innerHTML = DDG.exec_template('$template_name', $json_string);"
-								. "});</script>")->guts);
+					if(exists $structured->{templates}){ # user-specified templates
+						push @calls_goodie, $structured;
+						last;
+					}
+					else{ # auto-template
```
moollaza: @zachthompson aren't we're already using `DDG.duckbar()` to render the existing auto-templated Goodies?

Wondering if we can also use your `@calls_goodie` for these Goodie's as well? That way we can let the frontend deal with rendering rather than us shoving it into the page, which causes some problems (i.e #126)

PR #182, opened 2014-11-24
README.md updates for the new DuckDuckHack VM

mikedep333:
Hi. I am working with Jaryd on these updates. I will discuss them with him.

**The commit that changes the VM's username should not be merged until Jaryd (or someone else) accepts the new VM that I have been working on.**

PR #214, opened 2015-01-28
Update VMs: 2014-12-13 -> 2014-12-23

mikedep333:
Jaryd just uploaded these VMs.
As discussed in #182, the changes are:
1. Boot the VM automatically if the VM was improperly shutdown (important for those in headless mode)
2. Enable autologin to xfce4 desktop


moollaza:
@mikedep333 LGTM :+1: Thanks!

PR #223, opened 2015-02-10
Custom templates for Goodies

zachthompson:
Adds support for using custom templates with Goodies.  It has been tested with multiple templates and Handlebars helpers (see [DDG::Goodie::GoodieCustomTemplate](https://github.com/duckduckgo/zeroclickinfo-goodies/compare/zach/gtemplatev3) for a basic example; also intermittently available for viewing [here](http://zach.duckduckgo.com:5000/?q=gtc+great!&ia=software).)

There is an issue with templates not being available unless an artificial delay is added to the DDH.add call.  It's being evaluated internally so we may want to wait for the fix.  However, duckpan.js is already doing a similar thing for spices so it might not matter.

### DIFF
--- a/lib/App/DuckPAN/Web.pm
+++ b/lib/App/DuckPAN/Web.pm
@@ -272,7 +272,7 @@ sub request {
 					} elsif ($name =~ /handlebars$/){
 						$name =~ s/\.handlebars//;
 						$calls_template{$ia_name}{$name}{"content"} = $_;
-						$calls_template{$ia_name}{$name}{"is_ct_self"} = $result->call_type eq 'self';
+						$calls_template{$ia_name}{$name}{"is_ct_self"} = $is_goodie ? 1 : $result->call_type eq 'self';
 					}
 				}
 				push (@calls_nrj, $result->call_path) if ($result->can('call_path'));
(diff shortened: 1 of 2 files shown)


lib/App/DuckPAN/Web.pm
```diff
@@ -359,19 +359,17 @@ sub request {
...
-		my ($calls_nrj, $calls_script);
+		my $calls_nrj;
+		my $calls_script = join('', map { q|<script type='text/JavaScript' src='| . $_ . q|'></script>| } @calls_script);
 		# For now we only allow a single goodie. If that changes, we will do the
 		# same join/map as with spices.
 		if(@calls_goodie){
 			my $goodie = shift @calls_goodie;
 			$calls_nrj = "DDG.duckbar.future_signal_tab({signal:'high',from:'$goodie->{id}'});",
-			$calls_script = q|<script type="text/JavaScript" class="script-run-on-ready">/*DDH.add(| . encode_json($goodie) . q|);*/</script>|
+			$calls_script .= q|<script type="text/JavaScript" class="script-run-on-ready">/*window.setTimeout(DDH.add.bind(DDH, | . encode_json($goodie) . q|), 100);*/</script>|;
```
moollaza: @zachthompson please add a `#TODO` comment so we know that eventually the delay should be removed if/when it's handled in Core.

moollaza:
Tested and working!

Issue #225, opened 2015-02-12
VMWare image "duckpan upgrade" fails


tommytommytommy:
Running "duckpan upgrade" on the VMWare image from https://duck.co/duckduckhack/installing_duckpan gave me the following error message, and I was unable to force it to install the missing dependency using the options available for <code>duckpan upgrade</code> (probably because I just don't know how to?)

<pre>You already have latest version of DDG with 0.131

--> Working on http://duckpan.org/authors/id/M/MO/MOOLLAZA/App-DuckPAN-0.175.tar.gz
Fetching http://duckpan.org/authors/id/M/MO/MOOLLAZA/App-DuckPAN-0.175.tar.gz ... OK
Configuring App-DuckPAN-0.175 ... OK
==> Found dependencies: autodie
--> Working on autodie
Fetching http://www.cpan.org/authors/id/N/NT/NTHYKIER/autodie-2.26.tar.gz ... OK
Configuring autodie-2.26 ... OK
Building and testing autodie-2.26 ... FAIL
! Installing autodie failed. See /home/dax/.cpanm/work/1423741362.12939/build.log for details. Retry with --force to force install it.
! Installing the dependencies failed: Installed version (2.20) of autodie is not in range '2.25'
! Bailing out the installation for App-DuckPAN-0.175.
</pre>

tommytommytommy:
Here's the build log:

<pre>
cpanm (App::cpanminus) 1.6937 on perl 5.016003 built for x86_64-linux
Work directory is /home/dax/.cpanm/work/1423741362.12939
You have make /usr/bin/make
You have LWP 6.09
You have /bin/tar: tar (GNU tar) 1.26
Copyright (C) 2011 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by John Gilmore and Jay Fenlason.
--> Working on http://duckpan.org/authors/id/M/MO/MOOLLAZA/App-DuckPAN-0.175.tar.gz
Fetching http://duckpan.org/authors/id/M/MO/MOOLLAZA/App-DuckPAN-0.175.tar.gz
-> OK
Unpacking App-DuckPAN-0.175.tar.gz
Entering App-DuckPAN-0.175
Checking configure dependencies from META.json
[...]

PR #228, opened 2015-02-17
Update README.md to clarify which VM to download

mikedep333:
A user made a mistake in issue #225 by downloading the VirtualBox VM for VMware. I think that by moving the URL closer to the section header, we can minimize the likelihood of that mistake.


moollaza:
LGTM :+1:  Thanks a lot @mikedep333! Really appreciate it :smiley:

PR #230, opened 2015-02-19
Allow grabbing assets from sharedir sub-directories 

### LINKED ISSUE #136: Accessing directories in the share dir doesn't work
I have something like:

```javascript
DDG.get_asset_path('currency', 'assets/' + (DDG.is3x ? '96' : DDG.is2x ? '64' : '32') + '/' + symbol + '.png');
```

But it doesn't work on DuckPAN.

moollaza:
Fixes #136 

Explanation of changes here: https://github.com/duckduckgo/p5-app-duckpan/commit/021f3a2bc7792b83eef5512d17860510ba6c1b08

/cc @jagtalon @killerfish @mintsoft @MrChrisW @mattr555

### DIFF
--- a/lib/App/DuckPAN/Web.pm
+++ b/lib/App/DuckPAN/Web.pm
@@ -129,6 +129,7 @@ sub request {
 			$body = $errormsg;
 		}
 	} elsif (@path_parts && $path_parts[0] eq 'js' && $path_parts[1] eq 'spice') {
+		my $rewrite;
 		for (keys %{$self->_path_hash}) {
 			if ($request->request_uri =~ m/^$_/g) {
 				my $path_remainder = $request->request_uri;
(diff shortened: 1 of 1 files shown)


moollaza:
**Before:**

2015-02-18_23h50_04

**After:**
2015-02-18_23h50_31

**Note:** Not all flags are loaded from the share dir, only a few are.

mintsoft:
@moollaza will this be consumable from the images role too?

moollaza:
@mintsoft I haven't tested it but I don't think this should affect the images role. It goes about accessing the images via the `share` fn which should work properly -- does it not right now?

FWIW we're extremely close to launching the Goodies templates (technically the required changes are merged) which means we might not need the Images role, or it will need to be updated to specify a link to the asset like we use `DDG.get_asset_path` for

moollaza:
Seems that images role is unaffected by this update.

I think it's good for merge!

PR #233, opened 2015-02-28
Proper handling of UTF-8 content in _print_msg

### LINKED ISSUE #232: UTF8 Bug in _print_msg
When testing the Braille Goodie I noticed this:

2015-02-27_20h35_50

That error seems to be coming from the `_print_msg` [function](https://github.com/duckduckgo/p5-app-duckpan/blob/master/lib/App/DuckPAN.pm#L351)

moollaza:
Fixes #232

Before:

2015-02-27_20h35_50

After:

2015-02-27_20h58_58

/cc @zachthompson

### DIFF
--- a/lib/App/DuckPAN.pm
+++ b/lib/App/DuckPAN.pm
@@ -23,6 +23,7 @@ use Carp;
 use Encode;
 use Perl::Version;
 use Path::Tiny;
+use open qw/:std :utf8/;
 use App::DuckPAN::Cmd::Help;
 
 option dukgo_login => (
--- a/lib/App/DuckPAN/Cmd/Server.pm
+++ b/lib/App/DuckPAN/Cmd/Server.pm
@@ -12,7 +12,6 @@ use Path::Tiny;
 use LWP::Simple;
 use HTML::TreeBuilder;
 use Config::INI;
-use Data::Printer;
 use Term::ProgressBar;
 
 option port => (
--- a/lib/App/DuckPAN/Cmd/Server.pm
+++ b/lib/App/DuckPAN/Cmd/Server.pm
@@ -14,6 +14,7 @@ use HTML::TreeBuilder;
 use Config::INI;
 use Data::Printer;
 use Term::ProgressBar;
+use open qw/:std :utf8/;
 
 option port => (
     is => 'ro',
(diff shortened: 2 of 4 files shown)


moollaza:
Hmm actually this might not be right:

2015-02-27_21h02_41

I'm not sure if this a problem with `Data::Printer` our our code though. The frontend result is correct:

zachthompson:
If I add 

    use open qw/:std :utf8/;

to App::DuckPAN::Web/Query, both of those work for me.

zachthompson:
You can also just add it to App::DuckPAN, though it will involve more testing.

moollaza:
@zachthompson hmm it seems to work better for sure, but still some weird stuff happens for edge cases like `braille Mönchengladbach` -- that still spits out the broken text instead of the O (umlaut).

Also, it seems like `use open` needs to be added to App/DuckPAN.pm because otherwise the Braille IA throws an error that there's a wide character in print coming from `_print_msg`

zachthompson:
@moollaza I don't see broken text for "braille Mönchengladbach".  How can I reproduce?  I take it you're running server, as I don't think query accepts utf8.  If you add the use statement to both Query and Web, all of the filehandles passed to App::DuckPAN will be utf8.

moollaza:
@zachthompson sorry for the confusion, I was referring to DuckPAN Query -- DuckPAN Server works fine, and query no longer croaks, but I sometimes don't see characters rendered properly.

It's obviously not as important as the Server output, but I still use Query every now and then, so I was hoping I could get it to display UTF8 properly. I think it may also have to do with my terminal emulator and font so, there's clearly lots of variables at play.

@zachthompson if you're happy with the PR we should be good to merge then.

zachthompson:
Might want to add the use to App::DuckPAN::Query for Data::Printer.  Does A/D/C/Server.pm really need the use for the progress bar?  On a side note, I don't think A::D::C::Server is using Data::Printer.

moollaza:
@zachthompson whoops, meant to add it to Query.pm, not Server.pm! (Web.pm covers Server's use of Data::Printer)

>Does A/D/C/Server.pm really need the use for the progress bar?

Yup, it uses the progress bar [on initialization](https://github.com/duckduckgo/p5-app-duckpan/blob/zaahir/utf8-fix/lib/App/DuckPAN/Cmd/Server.pm) when it downloads the various files it requires from DDG.

For those of us on faster internet connections you barely notice it, but a previous dev had a really slow connection and was running into problems so he wanted to be able to check if things were completing and I guess how much longer they were going to take.

moollaza:
@zachthompson should be good now!

zachthompson:
I just meant does the progress bar need the "use open" as it's the only filehandle in the module.  But it looks like you fixed that. Looks good to me.

PR #235, opened 2015-03-02
Allow empty (self-closing) tags in answers.

### LINKED ISSUE #234: Empty tags are misinterpreted in answers
If an answer contains empty (self-closing) tags, like `<text />`, duckpan will output different markup via its web server. It will view the empty tag as a beginning one and by that possibly gain a sibling tag as child. This resulted in a bug in a SVG diagram that was present in production but not reproducible via duckpan.
`duckpan query` still outputs the original and correct markup.

This happens in the HTML::TreeBuilder module which uses the HTML::Parser which doesn't recognize empty tags if not specifically asked to. (Thanks for helping me pinpoint where this was happening, @moollaza)

SoerenWeber:
This should fix #234 by specifically asking HTML::Parser to recognize empty tags.

### DIFF
--- a/lib/App/DuckPAN/Web.pm
+++ b/lib/App/DuckPAN/Web.pm
@@ -341,6 +341,8 @@ sub request {
 						my $tb = HTML::TreeBuilder->new();
 						# Specifically allow unknown tags to support <svg> and <canvas>
 						$tb->ignore_unknown(0);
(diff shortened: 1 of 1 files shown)


moollaza:
LGTM :+1:

moollaza:
@SoerenWeber thanks again!

PR #237, opened 2015-03-11
Make Web.pm find the correct share dir for IAs that are substring of oth...

pfirsichbluete:
In duckduckgo/zeroclickinfo-spice the Time spice's name is a substring
of the Timer spice's name. When looking for resources for the Timer
IA, paths of the form

  .../timer/...

got mangled to

  .../time/r/...

because the matcher for IAs share directories did not check for the
trailing path separator. Hence, the share path of the Time (no trailing
“r” in Time) matched for Timer (trailing “r” in Timer) resources, and
path handling got confused.

By making the guard for identifying the relevant the share directory
also check on the trailing '/', the share directories of Time and
Timer are properly matched and the Timer IA again properly loads for
'timer 10 seconds'.

Fixes duckduckgo/zeroclickinfo-spice#1625

### DIFF
--- a/lib/App/DuckPAN/Web.pm
+++ b/lib/App/DuckPAN/Web.pm
@@ -87,7 +87,7 @@ sub request {
 	} elsif (@path_parts && $path_parts[0] eq 'share') {
 		my $share_dir;
 		for (keys %{$self->_share_dir_hash}) {
-			if ($request->path =~ m|^/$_|g) {
+			if ($request->path =~ m|^/$_/|g) {
 
 				$share_dir = $_;
 				my $filename = pop @path_parts;


moollaza:
@pfirsichbluete good catch! And thanks for taking the time to fix this.

I believe the same bug exists on line 134 as well? https://github.com/duckduckgo/p5-app-duckpan/blob/master/lib/App/DuckPAN/Web.pm#L134

pfirsichbluete:
> I believe the same bug exists on line 134 as well?

I could not reproduce on my end.

For me, the keys of `$self->_path_hash` (the relevant hash of line 134) already end in a `/`. Like

* `/js/spice/mtg/`
* `/js/spice/astrobin/fetch_id/`
* [...]

(while the keys of `$self->_share_dir_hash` did not).

Hence, the `$_` of
```
if ($request->request_uri =~ m/^$_/g) {
```
already contains the needed trailing `/` and is non susceptive to the bug.

Can you trigger the issue for line 134?

killerfish:
great work @pfirsichbluete I was just about to submit a pr for this :smile: 
> the keys of $self->_path_hash (the relevant hash of line 134) already end in a /

Yup, we shouldnt have an issue with that hash.

moollaza:
@pfirsichbluete yes, my mistake you're correct. Tested and LGTM :+1:

Issue #238, opened 2015-03-13
CSS containing UTF8 characters causes problems


moollaza:
While testing the BrainyQuote Spice (which uses a UTF8 quote character in the CSS) I noticed this error which prevents the CSS from loading.

![2015-03-13_13h28_03](https://cloud.githubusercontent.com/assets/873785/6643698/cef0fdd6-c984-11e4-8cb5-7fb7a71e2ca7.png)

moollaza:
@zachthompson I tried removing the change we made to Web.pm last week, `use open ...` and the problem went away.

Any ideas why this might be happening?

PR #239, opened 2015-03-14
Send page as utf-8

zachthompson:
Encode the body as utf-8 before adding to response (issue https://github.com/duckduckgo/p5-app-duckpan/issues/238).


moollaza:
@zachthompson I'm not sure this is correct -- or we're potentially double-encoding the UTF8 chars because this breaks the output:

2015-03-14_19h41_10
2015-03-14_19h42_34

zachthompson:
Right, we are decoding/re-encoding the spice api responses.

zachthompson:
Let's revisit this later.

PR #240, opened 2015-03-17
Die if auto-templating for goodies with >=3 items is requested

pfirsichbluete:
Currently, auto-templating is only supported for up to 2 items.

This has been pointed out by @zachthompson in:
  https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1055#discussion_r26614782

Not sure if such a warning is over the top or not.
It'd certainly help while developing.

### DIFF
--- a/lib/App/DuckPAN/Web.pm
+++ b/lib/App/DuckPAN/Web.pm
@@ -317,7 +317,10 @@ sub request {
 						last;
 					}
 					else{ # auto-template
-						my $template_name = 'goodie_'.scalar @{$structured->{input}}.'_inputs';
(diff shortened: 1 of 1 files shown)


moollaza:
LGTM :+1: 

@zachthompson any thoughts?

zachthompson:
@moollaza @pfirsichbluete What about just die'ing?  The template won't be there so not much point in continuing.

pfirsichbluete:
> What about just die'ing?

Done in 9f6c618. PR updated.

moollaza:
LGTM :+1:

PR #244, opened 2015-03-23
remove travisci dzil plugin, allow for faster container builds

moollaza:
/cc @zachthompson @jagtalon

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 ---
+sudo: false
 before_install:
   - export HARNESS_OPTIONS=j10:c HARNESS_TIMER=1
   - rm .travis.yml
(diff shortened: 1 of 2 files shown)


moollaza:
Doesn't shave off much time (we're not testing nearly enough of DuckPAN's functionality!), but I'm in favor of simplifying and removing dependencies.

.travis.yml
```diff
@@ -1,4 +1,5 @@
 ---
+sudo: false
 before_install:
   - export HARNESS_OPTIONS=j10:c HARNESS_TIMER=1
   - rm .travis.yml
```
zachthompson: I don't think we need to do this now.

moollaza: Right!

moollaza:
@zachthompson this should be good now :) -- thanks for catching that.

zachthompson:
Fantastic, off we go!

PR #249, opened 2015-04-27
Remove unneeded deps

zachthompson:
The CPAN::Repository dependency is causing failures since Dist::Data is no longer available.  Running tests, I don't see that it's actually needed.

Switch to AutoPrereqs which pulled the right deps as far as I could tell.

Note: the travis build error is from it trying to install the old duckpan which **this** PR should fix!


moollaza:
LGTM :+1:

PR #252, opened 2015-04-28
Add notice about Node & Handlebars.js requirements

moollaza:
Also, small fix to Codio image position.

/cc @zachthompson

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ We run our DuckPAN tests against Perl 5.16 and 5.18 using Travis (https://travis
 - [Vagrant Virtual Environment](#vagrant-virtual-environment)
     + [Setup Instructions](#setup-instructions)
 - [Installing DuckPAN Locally](#installing-duckpan-locally)
+    + [Optional Dependecies](#optional-dependecies-for-staff-and-maintainers)
     + [Adding Shell Completion](#adding-shell-completion)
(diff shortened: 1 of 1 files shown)


zachthompson:
@moollaza We should only need handlebars when running dzil in an IA repo.  Is the average duckpan user going to be doing this?

moollaza:
@zachthompson I think you'll need all the new dependencies if/when you try to run `duckpan test` -- I'm not sure how many devs actually use that vs `prove -Ili t/filename.t` though.

Maybe we should detect if they don't have those deps installed when they run something which needs them and give a message explainng what's needed? or we can refer them to these docs -- I can move these instructions to an "optional dependencies" section of sorts?

zachthompson:
An optional dependencies section would probably be more accurate.  Hopefully running things like test and release externally is rare. I like the "less is more" approach with respect to dependencies.

moollaza:
>I like the "less is more" approach with respect to dependencies.

100% with you on that!

moollaza:
@zachthompson I've changed things a bit, added an optional dependencies section which explains what's required and where it's required (i.e. dzil build/install/test)

moollaza:
In case someone runs into problems we can at least reference this.

zachthompson:
Looks good :+1:

PR #254, opened 2015-05-09
DuckPAN Test: default to prove, add --build for dzil test

moollaza:
This ensures that any users who run duckpan test won't run into dzil dependency issues (e.g. handlebars, etc)

Partially addresses #254 

/cc @zachthompson

### DIFF
--- a/lib/App/DuckPAN/Cmd/Test.pm
+++ b/lib/App/DuckPAN/Cmd/Test.pm
@@ -6,16 +6,24 @@ with qw( App::DuckPAN::Cmd );
 
 use MooX::Options protect_argv => 0;
 
+option full => (
+	is          => 'ro',
+	lazy        => 1,
+	short       => 'f',
+	default     => sub { 0 },
+	doc         => 'run full test suite via dzil',
+);
+
 sub run {
     my ( $self ) = @_;
 
     my $ret = 0;
 
-    if (-e 'dist.ini') {
-      $ret = system('dzil test');
-      $self->app->emit_error('Could not begin testing. Is Dist::Zilla installed?') if $ret == -1;
+    if ($self->full) {
+        $self->app->emit_error("Could not find dist.ini.") unless -e 'dist.ini';
+        $ret = system('dzil test');
+        $self->app->emit_error('Could not begin testing. Is Dist::Zilla installed?') if $ret == -1;
     } else {
-      $self->app->emit_notice("Could not find dist.ini.");
       $ret = system('prove -Ilib');
     }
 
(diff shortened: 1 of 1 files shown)


lib/App/DuckPAN/Cmd/Test.pm
```diff
@@ -6,16 +6,24 @@ with qw( App::DuckPAN::Cmd );
...
 
     my $ret = 0;
 
-    if (-e 'dist.ini') {
-      $ret = system('dzil test');
-      $self->app->emit_error('Could not begin testing. Is Dist::Zilla installed?') if $ret == -1;
+    if ($self->full) {
+        $self->app->emit_error("Could not find dist.ini.") unless -e 'dist.ini';
+        $ret = system('dzil test');
+        $self->app->emit_error('Could not begin testing. Is Dist::Zilla installed?') if $ret == -1;
```
zachthompson: We should just replace $ret == -1 with line 24.  Any non-zero exit value is an error.

On line 27 as well we should output an error if system fails.

moollaza: > We should just replace $ret == -1 with line 24. Any non-zero exit value is an error.

@zachthompson I think we still want to capture the return value though so we can `return` it below?

moollaza: > On line 27 as well we should output an error if system fails.

:+1:

moollaza:
@zachthompson I added in the error message when prove fails and simplified the check for a non-zero exist status.

Let me know if you think anything should be changed!

lib/App/DuckPAN/Cmd/Test.pm
```diff
@@ -6,20 +6,29 @@ with qw( App::DuckPAN::Cmd );
...
-      $ret = system('dzil test');
-      $self->app->emit_error('Could not begin testing. Is Dist::Zilla installed?') if $ret == -1;
-    } else {
-      $self->app->emit_notice("Could not find dist.ini.");
-      $ret = system('prove -Ilib');
-    }
+	if ($self->full) {
+		$self->app->emit_error("Could not find dist.ini.") unless -e "dist.ini";
+		$ret = system("dzil test");
+		$self->app->emit_error("Could not begin testing. Is Dist::Zilla installed?") if $ret;
```
zachthompson: I was just saying that there's no reason to catch the value and then test it.:
``` Perl
$self->app->emit_error("Could not begin testing. Is Dist::Zilla installed?") if $ret = system("dzil test");
```
Same with below.  Not a big deal, just more succinct.

moollaza: :+1:

moollaza:
@zachthompson this should be good now!

PR #256, opened 2015-05-14
Version check fix

moollaza:
This ensures that DuckPAN doesn't throw a fatal error when a new version of DuckPAN exists, but the user has the local version pin set lower. Instead it warns that a newer version exists and suggests that you update the local version pin.

Also fixed indentation inconsistencies in DuckPAN.pm and Perl.pm

/cc @zachthompson

### DIFF
--- a/lib/App/DuckPAN/Perl.pm
+++ b/lib/App/DuckPAN/Perl.pm
@@ -123,7 +123,7 @@ sub duckpan_install {
                         "You don't have $package installed. Installing latest version ($duckpan_module_version)";
 			$install_it = 1;
 		} elsif ($pin_version) {
-			$self->app->emit_info("$package: $localver installed, $pin_version pin, $duckpan_module_version latest");
+			$self->app->emit_info("$package: $localver installed, $pin_version pinned, $duckpan_module_version latest");
 			if ($pin_version != $localver) {
 				#  We continue here, even if the version is larger than latest released,
 				#  on the premise that there might exist unreleased development versions.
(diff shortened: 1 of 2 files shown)


zachthompson:
Do we really have any compelling reason to exit if they're not running the latest DDG/DuckPAN?  I would prefer that, unless we require some minimum version to function properly, upgrading would always just be a recommendation.

lib/App/DuckPAN/Perl.pm
```diff
@@ -131,9 +131,14 @@ sub duckpan_install {
 					$reinstall  = 1;       # Let us roll back, if necessary. Multiple packages may confuse this, but little harm.
 					$install_it = 1;
 				} else {
-					$message    = 'Could not locate version ' . $pin_version . ' of  ' . $package;
-					$install_it = 0;
+					$message = "Could not locate version $pin_version of \'$package\'";
```
zachthompson: Can remove the escapes.

moollaza:
> Do we really have any compelling reason to exit if they're not running the latest DDG/DuckPAN?

Not really? A constant warning that there's a new version would probably be compelling enough. There may be a few edge cases though, such as the recent `Locale::Country` changes you made. If they don't upgrade, the Goodies/Spice's dependent on that code would not work.

zachthompson:
Cool.  Let's remove the exiting in those cases then.  Maybe change emit_error to emit_notice too?  Devs shouldn't have to deal with an app that won't run because of optional versions.

lib/App/DuckPAN/Perl.pm
moollaza: :+1:

PR #257, opened 2015-05-16
Fixed naming collision in duckpan assets, use https

moollaza:
This PR enables DuckPAN to work correctly when using a DDG Dev instance as the upstream source. It likely resolves some other unnoticed (or minor) bugs.

I noticed a few odd problems:

- http vs https returned different files
- the /share/duckpan.js file was colliding with the remote duckpan.js, causing the latter to be injected multiple times, causing the internal templates to be erased from DDG.templates
- we were loading duckduckhack_js twice

/cc @zachthompson


zachthompson:
How do I specify that I want to use a different url for assets again?

PR #259, opened 2015-05-18
Add check and prompt for missing IA Bundles

moollaza:
/cc @zachthompson


moollaza:
Still need to clean up indents, but this works...

moollaza:
Results in this when any IA Bundle isn't installed or is outdated (non-fatal error)

2015-05-18_18h47_44

moollaza:
New output:

2015-05-18_19h14_03

PR #266, opened 2015-08-01
Check DDG before installing bundles

zachthompson:
Adds DDG to list of packages to be checked when bundles are in play.  Currently doesn't do this and fails when installing a bundle directly or checking requirements.

Fixed a few bugs in DuckPAN::Perl, tried to clarify the what the versions really mean, and added a bit more error checking.

### DIFF
--- a/lib/App/DuckPAN.pm
+++ b/lib/App/DuckPAN.pm
@@ -292,13 +292,13 @@ sub execute {
 				elsif($m eq 'ddg' && $ddg){ next }
 				push @modules, $_;
 			}
-			elsif (m/^duckpan|update|(upgrade|reinstall)$/i) {
-				my $upgrade_reinstall = lc $1;
+			elsif (m/^duckpan|update|(upgrade|reinstall|latest)$/i) {
+				my ($all_modules, $reinstall_latest) = map { lc } ($1, $2);
 				$self->empty_cache unless $self->empty;
 				push @modules, 'App::DuckPAN';
-				if($upgrade_reinstall){
+				if($all_modules){
 					push @modules, 'DDG', map { "DDG::${_}Bundle::OpenSourceDuckDuckGo" } qw(Goodie Spice Fathead Longtail);
-					unshift @modules, 'reinstall' if $upgrade_reinstall eq 'reinstall';
+					unshift @modules, $reinstall_latest if $reinstall_latest; 
 				}
 			}
 			else {
--- a/lib/App/DuckPAN.pm
+++ b/lib/App/DuckPAN.pm
@@ -277,7 +277,7 @@ sub execute {
 				/^(ddg)$/i ||
 				/(opensourceduckduckgo)$/i ||
 				/^(goodie)$/i || /^(spice)$/i || /^(fathead)$/i || /^(longtail)$/i ||
-				/^ddg::/ ||
+				/^ddg::/i ||
 				/^app/i) {
 				my $m = lc $1;
 
--- a/lib/App/DuckPAN.pm
+++ b/lib/App/DuckPAN.pm
@@ -276,16 +276,23 @@ sub execute {
 				/^dist/i ||
 				/^(ddg)$/i ||
 				/(opensourceduckduckgo)$/i ||
+				/^(goodie)$/i || /^(spice)$/ || /^(fathead)$/ || /^(longtail)$/ ||
 				/^app/i) {
 				my $m = lc $1;
+
+				if($m eq 'goodie' or $m eq 'spice' or $m eq 'fathead' or $m eq 'longtail'){
+					$_ = 'DDG::' . ucfirst($m) . 'Bundle::OpenSourceDuckDuckGo';
+					$m = 'opensourceduckduckgo';
+				}
+
 				if($m eq 'opensourceduckduckgo' && !$ddg){
 					unshift @modules, 'DDG';
-					++$ddg;
+					$ddg = 1;
 				}
 				elsif($m eq 'ddg' && $ddg){ next }
 				push @modules, $_;
 			}
-			elsif ($_ =~ m/^duckpan|update|(upgrade|reinstall)$/i) {
+			elsif (m/^duckpan|update|(upgrade|reinstall)$/i) {
 				my $upgrade_reinstall = lc $1;
 				$self->empty_cache unless $self->empty;
 				push @modules, 'App::DuckPAN';
--- a/lib/App/DuckPAN.pm
+++ b/lib/App/DuckPAN.pm
@@ -277,6 +277,7 @@ sub execute {
 				/^(ddg)$/i ||
 				/(opensourceduckduckgo)$/i ||
 				/^(goodie)$/i || /^(spice)$/i || /^(fathead)$/i || /^(longtail)$/i ||
+				/^ddg::/ ||
 				/^app/i) {
 				my $m = lc $1;
 
--- a/lib/App/DuckPAN.pm
+++ b/lib/App/DuckPAN.pm
@@ -26,6 +26,8 @@ use Path::Tiny;
 use open qw/:std :utf8/;
 use App::DuckPAN::Cmd::Help;
 
+no warnings 'uninitialized';
+
 option dukgo_login => (
 	is => 'ro',
 	lazy => 1,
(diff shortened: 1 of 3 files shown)


lib/App/DuckPAN/Perl.pm
```diff
@@ -109,54 +111,73 @@ sub duckpan_install {
 		$sp =~ s/\:\:/_/g;
 
 		# special case: check for a pinned verison number
-		my $pin_version            = $ENV{$sp};
-		my $localver               = $self->get_local_version($package);
-		my $duckpan_module_version = version->parse($module->version);
-		my $duckpan_module_url     = $self->app->duckpan . 'authors/id/' . $module->distribution->pathname;
+		my $pinned_version		= $ENV{$sp};
```
moollaza: nit: the indentation here is a bit too much

lib/App/DuckPAN.pm
```diff
@@ -268,21 +270,31 @@ sub execute {
...
+		my (@modules, @left_args, $ddg);
 		for (@arr_args) {
-			if ($_ =~ /^www/i ||
-				$_ =~ /^dist/i ||
-				$_ =~ /^ddg/i ||
-				$_ =~ /^app/i) {
+			if (/^www/i ||
+				/^dist/i ||
+				/^(ddg)$/i ||
+				/(opensourceduckduckgo)$/i ||
```
moollaza: wondering if we should make aliases for the ia bundles so we don’t have to `duckpan DDG::SpiceBundle::OpenSourceDuckDuckGo`?

e.g. `duckpan Spice` does `duckpan DDG::SpiceBundle::OpenSourceDuckDuckGo`

lib/App/DuckPAN/Perl.pm
```diff
@@ -109,54 +111,73 @@ sub duckpan_install {
...
 		$self->app->emit_notice($message);
 		push @to_install, $duckpan_module_url if ($install_it && !(first { $_ eq $duckpan_module_url } @to_install));
 	}
 
-	return 0 unless @to_install;
-	unshift @to_install, '--reinstall' if ($reinstall);    # cpanm will do the actual forcing.
-	return system("cpanm " . join(" ", @to_install));
+	if(@to_install){
+		unshift @to_install, '--reinstall' if $reinstall;    # cpanm will do the actual forcing.
+		if(system "cpanm --fdfdfdfdfdf @to_install"){
```
moollaza: What's the `--fdfdfdfdfd` for?

zachthompson: You don't know that special cpanm switch? ;)

lib/App/DuckPAN.pm
zachthompson: good idea :+1:

lib/App/DuckPAN/Perl.pm
```diff
@@ -109,54 +111,73 @@ sub duckpan_install {
...
-		my $pin_version            = $ENV{$sp};
-		my $localver               = $self->get_local_version($package);
-		my $duckpan_module_version = version->parse($module->version);
-		my $duckpan_module_url     = $self->app->duckpan . 'authors/id/' . $module->distribution->pathname;
+		my $pinned_version      = $ENV{$sp};
+		my $installed_version   = $self->get_local_version($package);
+		my $latest_version      = version->parse($module->version);
+		my $duckpan_module_url  = $self->app->duckpan . 'authors/id/' . $module->distribution->pathname;
 
-		$localver ||= 1e-6 if ($pin_version); # a silly, but true, value if missing and we need to compare with pinned.
```
moollaza: I thin we still need this? Or we need a better way to handle the locally installed version. When I run `duckpan reinstall` now, I'm getting "[FATAL]  Failed to find version 9.999 of App::DuckPAN".

We set the version to `9.999` if it's a local build.

lib/App/DuckPAN.pm
```diff
@@ -268,21 +270,38 @@ sub execute {
...
+				/^(goodie)$/i || /^(spice)$/ || /^(fathead)$/ || /^(longtail)$/ ||
```
moollaza: any reason spice, fathead and longtail don't have the `/i`?

lib/App/DuckPAN/Perl.pm
zachthompson: This line actually doesn't prevent that.  When I reconsidered what _reinstall_ should do I wanted to limit it to literally reinstalling the installed version.  It doesn't quite get there because the same if block handles the "not installed" case as well but it at least defaults to it (line 125.)  So you're trying to reinstall a version that's essentially not installed.  That is a fatal error imo.  How could it possibly reinstall that?

If you replace DuckPAN.pm and Perl.pm (or install it), you should see the desired behavior.  We could also check for 9.999 in this block and skip it.  Except for updating the reinstall logic and needing to test, are we really going to be doing a lot of reinstalling with local builds?

moollaza: I guess the name of the function should really be "reset" not so much "reinstall". It's meant to get a DDG dev back to where prod is.

Right this moment, I'm in need of this function. I've been doing both Spice and Goodie builds (and testing this PR) and so my Spice, Goodie and DuckPAN repos are out of sync with prod. Now that I want to test a prod release of the IA's I need to "reinstall" the currently released repos and so I would typically run "duckpan reinstall" to get all my repos back to where prod is at.

Does that clarify things?

zachthompson: Not really.  So why can't you `duckpan upgrade` to do that?  Updating to the latest doesn't get you to prod because prod will often be pinned lower than latest.  The next time you upgrade you are mysteriously downgraded to pinned versions.  That is not good behavior.  Are you saying that you want to get the latest without updating your pins to match?

moollaza: I don't believe `duckpan upgrade` will work in that case because my `localver` will be higher than anything released?

>The next time you upgrade you are mysteriously downgraded to pinned versions. That is not good behavior.

Agreed that behaviour sucks and has bitten me. I keep getting downgraded whenever I run `duckpan server` now. We should kill that. 

> Are you saying that you want to get the latest without updating your pins to match?

I thought so, but it might not be necessary. I suppose I can just update my pins.

zachthompson: > I don't believe duckpan upgrade will work in that case because my localver will be higher than anything released?

You're talking about 9.999, right?  The only time that happens is when you're using a repo for lib somehow, e.g. developing duckpan.   We could think about something like `duckpan latest` but externally that's what `upgrade` does; internally it would just allow you to circumvent the pins which I don't think we want to do.

moollaza: >You're talking about 9.999, right? 

Nope. If I dzil install the repo because I'm developing in it then my installed version of the repo would be 1 higher than latest. So IIRC `duckpan upgrade` won't have any effect as the installed version is higher than the currently released/pinned version.

zachthompson: Ok, again, that's a special case.  In that case it *should* say you have a newer version and leave it alone.  What behavior are you expecting?  All of these are internal development edge cases imo.  If you're dzil installing you're going under the hood and should be able get out of it, e.g. use `cpanm -U` first on you non-existent version and then `duckpan upgrade`.

Like I said, we could do `duckpan latest`.  In this case it would be like a reinstall but it would try to install `pinned` or `latest` versions, ignoring whatever mess created during development, e.g. installed_version.

moollaza: > All of these are internal development edge cases imo. 

I totally agree, it's really only a "problem" I face, but it happens very often, so I need an easy way to "fix" it.

> Like I said, we could do duckpan latest. In this case it would be like a reinstall but it would try to install pinned or latest versions, ignoring whatever mess was created during development, e.g. installed_version.

That works for me :+1:

zachthompson:
`latest` command added which basically ignores installed versions.

lib/App/DuckPAN.pm
```diff
@@ -268,21 +270,38 @@ sub execute {
...
-		my @left_args;
+		my (@modules, @left_args, $ddg);
 		for (@arr_args) {
-			if ($_ =~ /^www/i ||
-				$_ =~ /^dist/i ||
-				$_ =~ /^ddg/i ||
-				$_ =~ /^app/i) {
+			if (/^www/i ||
+				/^dist/i ||
+				/^(ddg)$/i ||
```
moollaza: Whoops we need to also accept `^ddg::` so that we handle `duckpan DDG::Publisher`, right now that commands falls through to the `emit_and_exit` below.

zachthompson: Updated.

moollaza:
LGTM :+1:

PR #271, opened 2015-09-28
Web.pm: Use beta.duckduckgo.com instead of all the DDH machines.

jagtalon:
Since we don't put any of the PRs in the DDH machines anymore (except for our own use), it's best to point this to the beta server. Since the beta server has all of the API keys new and old, using beta.duckduckgo.com is advantageous.

For @moollaza


zachthompson:
Aw yeah, thanks Jag!

PR #272, opened 2015-10-01
DuckPAN: No bundles by default

zachthompson:
There is no requirement for IA bundles to be installed in the development environment.  This removes them from being automatically considered during requirements checking and when upgrading/reinstalling/latest.  They can still be installed and managed manually.

Also fixes a bug with the reinstall/latest commands.

### DIFF
--- a/lib/App/DuckPAN.pm
+++ b/lib/App/DuckPAN.pm
@@ -420,7 +420,6 @@ sub check_requirements {
                         $self->check_perl &&
                         $self->check_app_duckpan &&
                         $self->check_ddg &&
-                        $self->check_ia_bundles &&
                         $self->check_ssh &&
                         $self->check_git);
         }
--- a/lib/App/DuckPAN.pm
+++ b/lib/App/DuckPAN.pm
@@ -402,33 +402,6 @@ sub phrase_to_camel {
 	return join('', map { ucfirst $_; } (split /\s+/, $phrase));
 }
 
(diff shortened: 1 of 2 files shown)


lib/App/DuckPAN.pm
```diff
@@ -300,12 +300,12 @@ sub execute {
 				elsif($m eq 'ddg' && $ddg){ next }
 				push @modules, $_;
 			}
-			elsif (m/^duckpan|update|(upgrade|reinstall|latest)$/i) {
+			elsif (m/^duckpan|update|(upgrade|(reinstall|latest))$/i) {
```
zachthompson: I'm debating whether we really want to support both update (only DuckPAN) and upgrade (DuckPAN + DDG) any longer.  Wouldn't it make sense to do both with update and individually if specified on the command-line?

moollaza: @zachthompson I've seen some CLI's that support both. `upgrade` could be used to upgrade installed packages and `update` would only update the DuckPAN client.

OSX's `homebrew` handles it the same way, whereas `npm` has an `update <pkg>` command.

moollaza:
@zachthompson this LGTM.

I realize that this has really been the only thing informing me that a new bundle was released. It's a separate issue, but I think we should still have something in place that notifies us (or at least me?) of new IA Bundle releases. Even an email would suffice I suppose.

PR #274, opened 2015-10-05
Fix latest command

zachthompson:
When `latest` is specified, version becomes global, potentially passing the wrong version for the package being evaluated.


lib/App/DuckPAN/Perl.pm
```diff
@@ -128,7 +128,8 @@ sub duckpan_install {
 		if($reinstall || $latest || !$installed_version) {
 			# Prefer versions in the following order when (re)installing: installed, pinned, latest
 			# Latest ignores the installed version
-			my $version = $installed_version unless $latest;
+			my $version;
```
moollaza: :+1:

moollaza:
@zachthompson this LGTM :+1:

PR #275, opened 2015-10-06
* Add missing 'use strict; use warnings;' line

Manwar:
Hi,

I propose fix to CPANTS warning regarding missing 'use strict; use warnings;' line.

Please review it.
Many Thanks.

Best Regards,
Mohammad S Anwar

### DIFF
--- a/bin/duckpan
+++ b/bin/duckpan
@@ -2,6 +2,8 @@
 # PODNAME: duckpan
 # ABSTRACT: Command line tool for using the DuckPAN of DuckDuckGo
 
+use strict; use warnings;
+
 my @libs;
 
 BEGIN {
--- a/t/lib/DDG/Goodie/TwoShoes.pm
+++ b/t/lib/DDG/Goodie/TwoShoes.pm
@@ -1,5 +1,6 @@
 package DDG::Goodie::TwoShoes;
 
+use strict; use warnings;
 use DDG::Goodie;
 
 triggers start => //;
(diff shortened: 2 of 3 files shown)


moollaza:
@Manwar thanks, it looks like they suggest we add it to `Cmd::Query`, `Cmd::Release` and `Cmd::Test` as well?

http://cpants.cpanauthors.org/dist/App-DuckPAN/errors

moollaza:
Actually, I just realized we're not using it anywhere...

Manwar:
In my humble opinion, where ever MooX is imported, you don't need 'use strict; use warnings;' as MooX imports Moo which brings in these checks.

moollaza:
> In my humble opinion, where ever MooX is imported, you don't need 'use strict; use warnings;' as MooX imports Moo which brings in these checks.

Ah, well then I guess your fix covers everything.

Let's add `no warnings 'uninitialized';` and then this should be good to merge.

Manwar:
Where do you suggest, we should add "no warnings 'uninitialized'"?

zachthompson:
@Manwar The problem with allowing moox to do this behind the scenes is, if/when we remove it or use something else, we almost certainly won't remember to add strict/warnings.  Doesn't hurt to have them redundantly specified afaik.

Manwar:
@zachthompson I agree with you.

zachthompson:
Thanks, we'll get the others later.

PR #276, opened 2015-10-06
Fix metadata on serp

zachthompson:
* After we stopped adding methods for each of the metadata fields in `DDG::Meta::Data`, it broke in duckpan.  
* The call to `get_js` also changed.
* Also handle adding metadata for cheat sheets.


moollaza:
LGTM :+1:

PR #277, opened 2015-10-16
Fix IA name prompt

zachthompson:
Need a space after the prompt for an IA name with `duckpan new`


moollaza:
odd. I thought that was there already.

Anyways LGTM :+1:

PR #278, opened 2015-10-26
Fix printing of failed to load instance answers

srvsh:
`p()`'s behaviour has changed in `Data::Printer >= 0.36` resulting in some unintended output. Specifically, it shows text such as `HASH(0x97d39e8)` after printing the list of instant answers that failed to load when starting `duckpan server`.

 This pull request is to work around this change in behaviour.

<s>Call emit_notice() on the return value of np() to fix the following.

- emit_notice() was called on the return value of p() from Data::Printer. Since
  this was a hashref, a superfluous string like HASH(0x97d39e8) was printed after
  the list of modules.
- p() printed to STDERR while emit_notice() printed to STDOUT.</s>

### DIFF
--- a/lib/App/DuckPAN/DDG.pm
+++ b/lib/App/DuckPAN/DDG.pm
@@ -6,7 +6,7 @@ with 'App::DuckPAN::HasApp';
 
 use Module::Pluggable::Object;
 use Class::Load ':all';
-use Data::Printer;
+use Data::Printer return_value => 'dump';
 use List::Util qw (first);
 
 sub get_dukgo_user_pass {
--- a/lib/App/DuckPAN/Query.pm
+++ b/lib/App/DuckPAN/Query.pm
@@ -2,7 +2,7 @@ package App::DuckPAN::Query;
 # ABSTRACT: Main application/loop for duckpan query
 
 use Moo;
-use Data::Printer;
+use Data::Printer return_value => 'dump';
 use POE qw( Wheel::ReadLine );
 use Try::Tiny;
 use open qw/:std :utf8/;
--- a/lib/App/DuckPAN/Web.pm
+++ b/lib/App/DuckPAN/Web.pm
@@ -13,7 +13,7 @@ use Plack::MIME;
 use HTML::Entities;
 use HTML::TreeBuilder;
 use HTML::Element;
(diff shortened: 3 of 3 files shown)


moollaza:
@srvsh thanks a lot for this. Can you describe what scenario currently recreates the problem so I can see it, and test your solution?

srvsh:
@moollaza Thanks for your quick response!

This issue can be reproduced if some instant answers fail to load when starting `duckpan server`. For example, the instant answer `DDG::Goodie::ZappBrannigan` needs the `YAML::XS` module and it'll fail to load if that's not present.

To simulate it, I run `perl -MDevel::Hide=YAML::XS $(which duckpan) server`. The output will now have a list of modules that failed to load (printed by `Data::Printer::pp()`) followed by a line such as `[NOTICE] HASH(0x97d39e8)` (printed by `emit_notice()`).

The fix ensures that all the output is from `emit_notice()` and removes the superfluous line at the end.

zachthompson:
This is due to a ["BACKWARDS-INCOMPATIBLE CHANGE"](https://metacpan.org/release/Data-Printer) in Data::Printer v0.36.

moollaza:
@zachthompson Aaah thanks for pointing that out! @srvsh we have our modules pinned in production and on our dev boxes so this issues doesn't appear for us right now.

It looks like we'd need to upgrade to the newest `Data::Printer` though to access the "experimental" `np()` function. So I'm not sure this will work for us right now.

We've brought this to our dev team's attention. For now I'm think we should label this as tolerated and leave it open.

srvsh:
Thanks @zachthompson and @moollaza . That's a good point about `np()` not being available to everyone. I think passing [`return_value => 'dump'`](https://metacpan.org/pod/release/GARU/Data-Printer-0.36/lib/Data/Printer.pm#Return-Value) while importing `Data::Printer` can be a good compromise to allow the code to work with earlier and newer versions of the module. This brings back the earlier behaviour of `p()`.

There were 3 modules where it was imported which I've updated in the new commit. The return value of `p()` is not used in `Web.pm` so it wasn't affected, but I still changed it just in case it's used in the future.

zachthompson:
Thank you.

PR #280, opened 2015-11-12
Fix ia_name_separated template variable

srvsh:
It does it by guessing where to add spaces and which letters to capitalize. $entered_name is used as a base instead of $name since that should be more like what the user intends to name the instant answer.

Examples where it works well:
DuckDuckGo   -> Duck Duck Go # spaces before capitals
duck duck go -> Duck Duck Go # capitalize at word boundaries
perl 6       -> Perl 6       # ditto
iPad Apps    -> iPad Apps    # no change since spaces and capitals are already present
HTML 5       -> HTML 5       # ditto

Examples where it does not:
iPad         -> i Pad        # space added between 'i' and 'Pad'

### DIFF
--- a/lib/App/DuckPAN.pm
+++ b/lib/App/DuckPAN.pm
@@ -401,6 +401,23 @@ sub phrase_to_camel {
 	return join('', map { ucfirst $_; } (split /\s+/, $phrase));
 }
 
+# Guess the name of the instant answer from user input
+sub phrase_to_separated_name {
+    my ($self, $phrase) = @_;
+    my $separated_name = $phrase;
+
+    for ($separated_name) { # Set the context
+        # Capitalize words if there are no capital letters
+        s/\b([a-z])/\U$1/g         unless /[A-Z]/;
+        # Insert spaces at transitions to captial letters if there are no spaces
+        s/(?<=[^A-Z])(?=[A-Z])/ /g unless / /;
+        # Clear the package/path separators
+        s!(::|/)\s*! !g;
+    }
+
+    return $separated_name;
+}
+
 sub check_requirements {
         my ($self) = @_;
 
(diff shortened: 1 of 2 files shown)


moollaza:
@srvsh thanks for this.

In the case of `iPad` (single word, no spaces) how about we just apply `ucfirst` to the word first, and then add spaces between `[a-z][A-Z]`

```perl
#examples
duckpan -> Duckpan -> "Duckpan"
duckPAN -> DuckPAN -> "Duck PAN"
duckPan -> DuckPan -> "Duck Pan"
iPad -> IPad -> "IPad"
goto -> Goto -> "Goto"
goTo -> GoTo -> "Go To"
```

/cc @zachthompson for thoughts + review

zachthompson:
@srvsh we recently updated the templates to remove most metadata, including [name](https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1755/files#diff-7ce4c356d0200c61c9ccdb93d2eab6e2L13).  If you want to modify the PR to remove the separated_name generation, that would be good.  Let me know, otherwise we can close this.

srvsh:
@zachthompson  I wrote a (tentative) [new template for cheatsheets](https://github.com/srvsh/zeroclickinfo-goodies/commit/577c53ec1c60b45c3489892885b4adee0c99e045) where I felt this could be useful for the `name` field. That was the main reason I wanted to fix the issue with `ia_name_separated`. Do you think I should change the template to not use the name variables?

zachthompson:
Got it.  We want to remove those values from cheat sheets *eventually* but we're not quite ready to do that.  I'll give this a look too. Thanks.

moollaza:
@srvsh this has a merge conflict now --seeing as I just merged your other PR I think we can we close?

srvsh:
@moollaza Sure, this can be closed since this PR would only affect the `name` field of Instant Answers in a limited way. Thanks for taking a look at it.

PR #283, opened 2015-11-25
Add alt_to values to rewrites

zachthompson:
Use the new [alt_rewrites](https://github.com/duckduckgo/duckduckgo/pull/163) method to obtain the `alt_to` links, if they exist from the parent.

Depends on https://github.com/duckduckgo/duckduckgo/pull/163

### DIFF
--- a/lib/App/DuckPAN/Web.pm
+++ b/lib/App/DuckPAN/Web.pm
@@ -63,8 +63,15 @@ sub BUILD {
 			}
 			$share_dir_hash{$_->module_share_dir} = ref $_ if $_->can('module_share_dir');
 			$path_hash{$_->path} = ref $_ if $_->can('path');
+
+			my $alt_rewrites = $_->alt_rewrites;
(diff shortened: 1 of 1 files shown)


lib/App/DuckPAN/Web.pm
```diff
@@ -63,8 +63,15 @@ sub BUILD {
 			}
 			$share_dir_hash{$_->module_share_dir} = ref $_ if $_->can('module_share_dir');
 			$path_hash{$_->path} = ref $_ if $_->can('path');
+
+			my $alt_rewrites = $_->alt_rewrites;
```
moollaza: I think this should be inside the `if ($_->does('DDG::IsSpice')) {` block? Only Spice will have the `alt_to`

zachthompson: Yes! The while loop too.

moollaza:
LGTM -- Tested and working :+1:

PR #287, opened 2015-12-08
Capture rewrites before loop

zachthompson:
Get all of the rewrites in a var before loop


moollaza:
Tested and working :+1:

PR #289, opened 2015-12-09
New.pm: add ia_id template variable

srvsh:
... and remove ia_name and lia_name.

Tested with changes to Goodie and Spice packages.


moollaza:
@srvsh Thanks! This LGTM :+1:

PR #279, opened 2015-11-09
Support for generic templates

srvsh:
1. Moved template information out of `p5-app-duckpan` and into respective IA packages
2. The templates are now defined in YAML files. See https://github.com/srvsh/zeroclickinfo-goodies/blob/3a4f248f61cdb343df37f0d84861129d91c9de7f/template/templates.yml for an example. The file also has some documentation about the structure of the YAML file.
3. An instant answer can have multiple templates that can be chosen by the user during creation.
4. A template has some required files and some optional files. Optional files are in groups and users must specify which of these groups of files they'd like to generate.
3. `App::DuckPAN::TemplateDefinitions` is responsible for reading the YAML file and creating one `App::DuckPAN::Template` instance for each template. 
4. `App::DuckPAN::Template` is responsible for creating output files based on variables provided and a list of optional file groups. It also provides a list of output directories that `App::DuckPAN::Restart` watches for changes.

Usage:
duckpan new "Hello World" # creates an instant answer with the 'default' template
duckpan new --template cheatsheet "Hello World" # creates an instant answer with the 'cheatsheet' template

TODO:
1. <s>Tests</s>
2. <s>An option to list all available templates</s>
3. <s>Prompting the user to update the instant answer repository if a `templates.yml` file is not found</s>

(this is part of work related to Issue #268)

PR #294, opened 2015-12-22
Restart.pm: workaround for missing templates.yml

srvsh:
If the latest `duckpan` app is used with an older IA repository, running `duckpan server` can fail if the templates definitions file, `templates.yml`, is missing in the IA repository. This file is used to get a list of directories that need to be monitored for changes by the restarter. After the failure, the restarter then restarts the server and this process repeats forever. (See [this comment](https://github.com/duckduckgo/p5-app-duckpan/pull/279#issuecomment-166473849) for details)

With this workaround, if `templates.yml` is missing, the [older method](https://github.com/duckduckgo/p5-app-duckpan/blob/13382bd13596f3264de3cc8581006a5838d3f1fd/lib/App/DuckPAN/Restart.pm#L70-L76) of determining directories to monitor is used, thus allowing the server to run correctly.

### DIFF
--- a/lib/App/DuckPAN/Restart.pm
+++ b/lib/App/DuckPAN/Restart.pm
@@ -5,6 +5,7 @@ use File::Find::Rule;
 use Filesys::Notify::Simple;
 
 use App::DuckPAN::TemplateDefinitions;
+use Try::Tiny;
 
 use strict;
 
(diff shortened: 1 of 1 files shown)


lib/App/DuckPAN/Restart.pm
```diff
@@ -58,31 +59,65 @@ sub run_restarter {
...
+            # There was a problem loading the template definitions file. This
+            # can happen if the instant answer repository is of an older
+            # version and does not have the templates.yml file.
+            #
+            # In this case we use the older method of determining directories
+            # to monitor. This support can be removed sometime in the future.
+            # Note: Cheatsheet directories will not be monitored with this
+            # method.
+
+            $self->app->emit_debug("Instant Answers repository does not have a " .
```
moollaza: I'm not seeing this message anywhere but the code is definitely working because the server isn't breaking down :+1:

moollaza:
Would like @zachthompson to review this later, but for now I think this is fine :+1: 

Will test more and merge.

lib/App/DuckPAN/Restart.pm
srvsh: @moollaza Thanks, I have set the message to be shown in verbose mode because I think users don't need to be concerned about it normally. You should be able to see the message by running `duckpan -v server`.

moollaza: @srvsh thanks, I actually tried it with verbose mode and didn't notice it but will try again

moollaza:
@srvsh Thanks! This LGTM :+1:

srvsh:
Thanks!

PR #296, opened 2015-12-26
Restart.pm: Don't use the 'needs_restart' fields from templates.yml file

srvsh:
When adding support for generic templates, I had added a `needs_restart` field to the `templates.yml` files for [Goodie](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/template/templates.yml#L15) and [Spice](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/template/templates.yml#L14) IAs to indicate which files (actually directories) should cause the server to restart when they are changed. The problem with that is that the `share/goodie` and `share/spice` directories are no longer watched for changes because no template-generated files in those directories (handlebars, css, js) need the server to be restarted. This is not right though, since now changes to other custom files such as config or data files of IAs do no cause a restart as well.

This PR reverts the 'needs_restart' change  and uses the old way of determining which directories to watch - basically all directories which contain auto-generated files, except for 'test' directories. I'll submit a couple of other PRs, one each for Goodie and Spice repositories to remove the field from the `templates.yml` files as well.

Thanks, and sorry about the error.

Testing steps:

1. Run `duckpan server` from a Goodie or Spice repository.
2. Create a new file in `share/`. The server will not restart.
3. Create or modifiy a file in `share/goodie` or `share/spice` or any subdirectory. The server will restart if the file has an extension other than `handlebars`, `css` or `js`.
4. Create or modify a file in `t/`. The server will not restart.

### DIFF
--- a/lib/App/DuckPAN/Restart.pm
+++ b/lib/App/DuckPAN/Restart.pm
@@ -69,7 +69,7 @@ sub _get_directories_to_monitor {
 
         for my $template (@templates) {
             push @output_dirs, $template->output_directory
-                if $template->needs_restart;
+                unless $template->name =~ /test/;
         }
     } catch {
         if (/template definitions/i) {
--- a/lib/App/DuckPAN/Restart.pm
+++ b/lib/App/DuckPAN/Restart.pm
@@ -71,7 +71,8 @@ sub _get_directories_to_monitor {
             push @output_dirs, $template->output_directory
                 unless $template->name =~ /test/;
         }
-    } catch {
+    }
+    catch {
         if (/template definitions/i) {
             # There was a problem loading the template definitions file. This
             # can happen if the instant answer repository is of an older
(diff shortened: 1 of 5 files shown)


zachthompson:
Hey @srvsh, while we're here can we change `_get_directories_to_monitor` to [return a reference](https://github.com/duckduckgo/p5-app-duckpan/pull/296/files#diff-ec50fbebe34d400acf9cbb24be0ef470R107)?

lib/App/DuckPAN/Restart.pm
```diff
@@ -69,7 +69,7 @@ sub _get_directories_to_monitor {
 
         for my $template (@templates) {
             push @output_dirs, $template->output_directory
-                if $template->needs_restart;
+                unless $template->name =~ /test/;
         }
     } catch {
```
zachthompson: No cuddled catches/elses, please.

srvsh: Done

srvsh:
Thanks @zachthompson, I have made the change to return a reference instead of a list from `_get_directories_to_monitor()`.

zachthompson:
Thank you :+1:

PR #304, opened 2016-01-20
Add handle to template

zachthompson:
Allow developers to choose `handle` during the template process when creating a new IA.  Should go with associated [goodie](https://github.com/duckduckgo/zeroclickinfo-goodies/pull/2085) and [spice](https://github.com/duckduckgo/zeroclickinfo-spice/pull/2427) template PRs.

### DIFF
--- a/lib/App/DuckPAN/Cmd/New.pm
+++ b/lib/App/DuckPAN/Cmd/New.pm
@@ -196,6 +196,8 @@ sub run {
 		$lc_name = $lc_path . '_' . $lc_name;
 	}
 
+	# If the Perl module every becomes optional, this should only run if the user
+	# requests one
 	my $handler = $self->_config_handler;
 
 	my @optional_templates = $self->_ask_optional_templates
@@ -217,9 +219,9 @@ sub run {
 
 	# Show the list of files that were successfully created
 	my @created_files = @{$generate_result{created_files}};
-	$self->app->emit_info("Created files:");
+	$self->app->emit_info('Created files:');
 	$self->app->emit_info("    $_")     for    @created_files;
-	$self->app->emit_info("    (none)") unless @created_files; # possible on error
+	$self->app->emit_info('    (none)') unless @created_files; # possible on error
 
 	if (my $error = $generate_result{error}) {
 		# Remove the line number information if not in verbose mode.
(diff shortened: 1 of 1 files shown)


lib/App/DuckPAN/Cmd/New.pm
```diff
@@ -226,7 +233,39 @@ sub run {
...
+		'remainder: (default) The query without the trigger words, spacing and case are preserved.',
+		'query_raw: Like remainder but with trigger words intact',
+		'query: Full query normalized with a single space between terms',
+		'query_lc: Like query but in lowercase',
+		'query_clean: Like query_lc but with non-alphanumeric characters removed',
+		'query_nowhitespace: All whitespace removed',
+		'query_nowhitespace_nodash: All whitespace and hyphens removed',
+		'words: Like query_clean but returns an array of the terms split on whitespace',
+		'query_parts: Like query but returns an array of the terms split on whitespace',
+		'query_raw_parts: Like query_parts but array contains original whitespace elements'
```
moollaza: should we add "matches"? We'd need to indicate it only works with a `regex` triggers though...

zachthompson: Yes, we should.  I just spaced it.

russellholt:
:+1:

zachthompson:
`matches` has been added.  Since the default triggers wouldn't have worked with it, I added very basic trigger templating as well.  I didn't want to spend too much time on the trigger part since we'll probably be migrating them to metadata.  So there's no dialogue for them.

moollaza:
>  I didn't want to spend too much time on the trigger part since we'll probably be migrating them to metadata. So there's no dialogue for them.

Right! I forgot about that.

zachthompson:
@GuiltyDolphin This needs to go with those two handle PRs you merged.

GuiltyDolphin:
@zachthompson Whoops, forgetful me!

moollaza:
I meant to comment here that I think we should handle this more like the optional template components, by asking if you wish you configure it.

I'm worried that a new developer might now know what the `handler` is or why they would want to change it, so the 11 options may be overwhelming.

Or we should break it down into a couple steps:
 -> Do you need a regex trigger? (default is "Word" triggers) [y/N]
 -> Would you like to configure the `handle` function? (default handle is "remainder") [y/N]

--

The downside is that this is two more steps to the process...

PR #305, opened 2016-01-22
Fix not testing packages in directories

GuiltyDolphin:
`duckpan test` was not testing anything within a directory.

### DIFF
--- a/lib/App/DuckPAN/Cmd/Test.pm
+++ b/lib/App/DuckPAN/Cmd/Test.pm
@@ -23,7 +23,7 @@ sub run {
 		$self->app->emit_error("Could not find dist.ini.") unless -e "dist.ini";
 		$self->app->emit_error("Could not begin testing. Is Dist::Zilla installed?") if $ret = system("dzil test");
 	} else {
-		$self->app->emit_error("Tests failed! See output above for details") if $ret = system("prove -Ilib t/*");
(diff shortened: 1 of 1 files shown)


lib/App/DuckPAN/Cmd/Test.pm
```diff
@@ -23,7 +23,7 @@ sub run {
 		$self->app->emit_error("Could not find dist.ini.") unless -e "dist.ini";
 		$self->app->emit_error("Could not begin testing. Is Dist::Zilla installed?") if $ret = system("dzil test");
 	} else {
-		$self->app->emit_error("Tests failed! See output above for details") if $ret = system("prove -Ilib");
+		$self->app->emit_error("Tests failed! See output above for details") if $ret = system("prove -Ilib t/*");
```
zachthompson: I think we want:
``` Perl
system 'prove -lr t';
```
here to catch arbitrarily nested subdirs.

GuiltyDolphin: @zachthompson You're right.

I was thinking of adding tests to make sure directories work (and for regression) - I can't see any tests that are testing the existing `duckpan test`? Any pointers?

zachthompson: `prove -r` should take care of directories but if you think of other tests that might be valuable let us know.

zachthompson:
Taken care of in https://github.com/duckduckgo/p5-app-duckpan/pull/307

PR #308, opened 2016-01-22
Skip install when using dev versions

zachthompson:
Keeps installing `DDG` when using the repo:

* Make the `dev_version` an option
* Skip installing when we detect the `dev_version` for a package


moollaza:
Thanks! This LGTM :+1:

PR #312, opened 2016-02-09
Update css-rewrite regex for single and double quotes

moollaza:
/cc @zachthompson @andrey-p 

@andrey-p I wasn't getting any console errors about broken JS before, but I think this does solve a small bug.

What did you do to produce the error that led you to this function?


moollaza:
Spoke with @andrey-p and this fixed the problem he was seeing :+1:

PR #316, opened 2016-02-25
Fix position of attribution link

### LINKED ISSUE #292: DuckPAN renders (i) icon incorrectly
image

I'm not sure why it's so poorly positioned?

### LINKED ISSUE #310: DuckDuckGo CSS Loading: force order to mimic DuckDuckGo.com
After a while of seeing various oddities with the UI when testing in DuckPAN I've realized that the cause is that we're loading the CCS files we pull down from DuckDuckGo in a different order than how they're loaded in production.

Seeing as order matters in CSS, certain rules aren't overriding each other, causing some visual discrepancies.

We need to ensure the CSS are concatenated in the same order that they appear on DuckDuckGo. As a test, I manually dropped the current CSS links into a DuckPAN window and the UI corrected itself

Mailkov:
Fixes #310 & Fixes #292

### DIFF
--- a/lib/App/DuckPAN/Cmd/Server.pm
+++ b/lib/App/DuckPAN/Cmd/Server.pm
@@ -355,21 +355,26 @@ sub get_sub_assets {
 		}
 	}
 
+    my @cssfile;
 	for (grep { $_->attr('type') && $_->attr('type') eq 'text/css' } @link) {
 		if (my $href = $_->attr('href')) {
 			# We're looking for txxx.css and sxxx.css.
 			# style.css and static.css are for development mode.
 			if ($href =~ m/^\/((?:[str]\d+|style|static|serp)\.css)/) {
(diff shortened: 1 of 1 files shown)


lib/App/DuckPAN/Cmd/Server.pm
```diff
@@ -355,21 +355,26 @@ sub get_sub_assets {
...
-				  {
-					name     => $name . ' CSS',
-					internal => $cache_path->child($name),
-					external => $name
-				  };
+				push @cssfile, $name;
 			}
 		}
 	}
+    foreach (sort @cssfile) {
```
moollaza: Looks like there's mixed tabs & spaces?

This file uses hard-tabs, so lets stick with those

moollaza:
@Mailkov Thanks! This LGTM :+1:

PR #317, opened 2016-03-13
Allow user to use ID rather than name when testing

GuiltyDolphin:
So you could do `duckpan test html_entities_decode` rather than `duckpan test HTMLEntitiesDecode`.

Also supports cheat sheets (in Goodies); so you could do `duckpan test vim_cheat_sheet`. Assuming duckduckgo/zeroclickinfo-goodies#2740 goes through, multiple cheat sheets could be specified.

### To-Do

- [x] Add special support for Cheat Sheets (so you could do `duckpan test vim_cheat_sheet`).

### DIFF
--- a/lib/App/DuckPAN.pm
+++ b/lib/App/DuckPAN.pm
@@ -427,11 +427,17 @@ sub phrase_to_camel {
 	return $camel;
 }
 
-sub _ia_names {
+has _ia_names => (
+	is      => 'ro',
+	lazy    => 1,
+	builder => '_build_ia_names',
+);
+
+sub _build_ia_names {
 	my $self = shift;
 	my @test_paths = File::Find::Rule->name('*.t')->in('t');
 	my @names = map { scalar(fileparse($_, qr/\.[^.]*/)) } @test_paths;
-	return @names;
+	return \@names;
 }
 
 # Normalize an Instant Answer name to a standard form.
@@ -440,8 +446,8 @@ sub normalize_ia_name {
 	my ($self, $name) = @_;
 	$name =~ s/_//g;
 	$name = lc $self->phrase_to_camel($name);
-	my @known_ias = $self->_ia_names();
-	return first { lc $_ eq $name } @known_ias;
+	my $known_ias = $self->_ia_names();
+	return first { lc $_ eq $name } @$known_ias;
 }
 
 sub check_requirements {
--- a/lib/App/DuckPAN.pm
+++ b/lib/App/DuckPAN.pm
@@ -9,6 +9,7 @@ use MooX::Options;
 use App::DuckPAN::Config;
 use File::Which;
 use File::Basename;
+use File::Find::Rule;
 use Class::Load ':all';
 use HTTP::Request::Common qw( GET POST );
 use HTTP::Status;
--- a/lib/App/DuckPAN.pm
+++ b/lib/App/DuckPAN.pm
@@ -27,6 +27,7 @@ use Perl::Version;
 use Path::Tiny;
 use open qw/:std :utf8/;
 use App::DuckPAN::Cmd::Help;
+use DDG::Meta::Data;
 
 no warnings 'uninitialized';
 
@@ -428,27 +429,18 @@ sub phrase_to_camel {
 	return $camel;
 }
 
-has _ia_names => (
-	is      => 'ro',
-	lazy    => 1,
-	builder => '_build_ia_names',
-);
-
-sub _build_ia_names {
-	my $self = shift;
-	my @test_paths = File::Find::Rule->name('*.t')->in('t');
-	my @names = map { scalar(fileparse($_, qr/\.[^.]*/)) } @test_paths;
-	return \@names;
-}
-
 # Normalize an Instant Answer name to a standard form.
 # Returns undef if an IA matching the given name cannot be found.
 sub normalize_ia_name {
 	my ($self, $name) = @_;
-	$name =~ s/_//g;
-	$name = lc $self->phrase_to_camel($name);
-	my $known_ias = $self->_ia_names();
-	return first { lc $_ eq $name } @$known_ias;
+	my $ia = $name =~ /_/
+		? DDG::Meta::Data->get_ia(id => $name)
+		: DDG::Meta::Data->get_ia(id => $self->camel_to_underscore($name))
+		or $self->emit_and_exit(1, "No Instant Answer found with name '$name'");
+	my $mod_name = $ia->{perl_module};
+	$mod_name =~ /^DDG::(?:Goodie|Spice)(?:::[^:]+)*::([^:]+)$/
+		or $self->emit_and_exit(1, "Must be in the root of a checked-out Instant Answer repository to run DuckPAN");
+	return $1;
 }
 
 sub check_requirements {
--- a/lib/App/DuckPAN.pm
+++ b/lib/App/DuckPAN.pm
@@ -431,16 +431,21 @@ sub phrase_to_camel {
 
 # Normalize an Instant Answer name to a standard form.
 # Returns undef if an IA matching the given name cannot be found.
-sub normalize_ia_name {
+sub get_ia_by_name {
 	my ($self, $name) = @_;
-	my $ia = $name =~ /_/
-		? DDG::Meta::Data->get_ia(id => $name)
-		: DDG::Meta::Data->get_ia(id => $self->camel_to_underscore($name))
-		or $self->emit_and_exit(1, "No Instant Answer found with name '$name'");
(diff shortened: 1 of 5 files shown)


zachthompson:
If we go this route I think it would be better if we allowed module or id across the board so you could do `duckpan server cheat_sheets` for example.

GuiltyDolphin:
@zachthompson Hmm... I'll look into that. We could provide a standard name-conversion function, I reckon we could still use the tests folder as a fallback for name conversions though (some IAs have names that are hard to capture using regexes (e.g., where a component represents multiple upper-case characters).

zachthompson:
Exactly, a function we can use throughout to convert.  We probably want to use DDG::Meta::Data to do the decoding.

GuiltyDolphin:
@zachthompson Righty, I've added support for IDs (and lowercase etc.) in addition to standard names for any command using blocks.

I didn't go down the Meta route as I felt it would add a large amount of overhead we don't really need? Instead I'm checking the tests directory for all the IAs (AFAIK we are guaranteed that all IAs supported by DuckPAN have tests?). This also allows for the weird ID not mapping directly to IA name cases (`html_entities_decode` & `htmlentitiesdecode` -> `HTMLEntitiesDecode`).

I've tested it out a bunch and it seems to be working pretty well (I've noticed no impact on performance).

A nice addition might be (if we don't have it already) to provide a hash from these normalized names to appropriate files (tests, lib, share etc.) so that no commands need to do *any* normalizing.

zachthompson:
The metadata route doesn't add any overhead since it's already required and used.  In any event we want to move away from the module name and id having to be tied together.

GuiltyDolphin:
@zachthompson Herum, what would you recommend using then? It looks like we need an ID or package name to actually grab info from `Meta::Data`? Currently I'm just normalizing the name to a form where it equals a file in the tests directory when case-folded - so we are guaranteed to find it if it exists (doesn't have to be ID - can be anything that would 'normalize' to the name). Are we allowed filenames for IAs that are completely different to the name?? I thought Perl's package system didn't allow that?

zachthompson:
@GuiltyDolphin If an id is provided it can be looked up directly, otherwise you can look it up by module:
```Perl
my $ia = DDG::Meta::Data->get_ia(id => 'text');
or
my @ia = DDG::Meta::Data->get_ia(module => 'DDG::IA::CreateMe');
```
Will that suffice?

GuiltyDolphin:
@zachthompson Doesn't that mean module and ID is tied together? I think the current approach is much looser on what is allowed (`cheat sheets`, `cheat_sheets`, `CheatSheets` etc. would all work). Why is it we want to look up the full IA? Many of the commands don't need any more info than the name to work with (and most aren't using Blocks).

Any that are using Blocks automatically get the benefit of the additional name forms - others can call `normalize_ia_name` explicitly if they so wish.

I'm still not exactly sure what the advantage would be to using `Meta::Data` over what I've done currently.

zachthompson:
Because this relationship,
```
my_instant_answer <=> MyInstantAnswer
```
may not hold in the future.  In fact, it's something that we would like to get rid of because it causes problems and limits what we can do.  Just off the top of my head, you could probably do,
```Perl
my $pm = DDG::Meta::Data->get_ia(id => $id)->{perl_module};
unless($pm =~ /(?Goodie|Spice)::(.+)$/){
      die
}
my $t = "$1.t";
```
and it will work if we break that relationship in the future.  Frankly, I would much rather we did:
```Perl
my @t = DDG::Meta::Data->get_ia(id => $id)->{test_filles};
```
but it's just not available yet.

Same with the share directory and test file needing to match the id which ultimately could be anything as long as it was specified in metadata.

GuiltyDolphin:
@zachthompson Ohh! You mean to treat the *ID* as static and we should *only* use that when finding information about IAs? 

Also yeah, it'd be really nice to be able to just look up test and share files in that manner, I think I mentioned previously it might be something to do in the future to provide an interface to the directory structure we build (we currently have the templates system - maybe we could do something with that?).

zachthompson:
Yeah, it's best to not proliferate logic that does the camel case to underscore conversion.  If an id is supplied, it should just match without us having to manipulate anything.

There are quite few other things that would be best in metadata.   Is there an issue for the directory structure interface idea?

GuiltyDolphin:
@zachthompson I don't think so - just a previous comment.

I'll do a bit of digging and see if I can come up with some ideas to throw into an issue.

GuiltyDolphin:
@zachthompson `get_ia_by_id` is taking an awful long time to do anything, I just ran the DuckPAN tests while using it and they clocked in at 58 seconds. Wasn't the caching issue fixed a while back?

GuiltyDolphin:
@zachthompson With `use DDG::Meta::Data` in `DuckPAN.pm` it is taking an awful long time to do anything. Re-running the tests clocked in at 58 and 66 seconds - wasn't the caching issue supposed to be fixed a while back?

zachthompson:
I'm not sure what you're doing.  Nothing it does takes that long.  Where's the call you're making?

GuiltyDolphin:
`by_id`, but it isn't that that makes it take time, only the import.

I switched to my branch with the metadata caching addition on and it ran a lot faster - so it's definitely the caching.

zachthompson:
I'm not sure they've pushed the fix through but I don't see why you don't set `NO_METADATA_DOWNLOAD` and forget about it.

GuiltyDolphin:
@zachthompson Sure - I'll set it locally, I'm a bit worried about others in a similar situation though - I hadn't even heard of it before I looked through the code. Perhaps we should make it clear in the docs that it is available if needed? (Unless I'm missing something...)

GuiltyDolphin:
@zachthompson I've got a failing build, saying it can't find `DDG::Meta::Data` https://travis-ci.org/duckduckgo/p5-app-duckpan/jobs/120032109#L407.

We use `DDG::Meta::Data` elsewhere, so any idea what's going on?

zachthompson:
@GuiltyDolphin it's [this line](https://github.com/duckduckgo/p5-app-duckpan/blob/master/dist.ini#L39) in dist.ini.  I don't recall why we needed to do that but it appears to be fine [without it](https://travis-ci.org/duckduckgo/p5-app-duckpan/builds/120044269) now.  Try removing that line. You could also remove Perl 5.14/5.18 to speed things up as we don't need to test those.

GuiltyDolphin:
@zachthompson Seems to have done the trick! :+1:

By the way, if you're free on Slack I've got some questions about `$loc` and `$lang` (not related to this).

lib/App/DuckPAN.pm
```diff
@@ -426,6 +429,20 @@ sub phrase_to_camel {
...
+# Normalize an Instant Answer name to a standard form.
+# Returns undef if an IA matching the given name cannot be found.
+sub normalize_ia_name {
+	my ($self, $name) = @_;
+	my $ia = $name =~ /_/
+		? DDG::Meta::Data->get_ia(id => $name)
+		: DDG::Meta::Data->get_ia(id => $self->camel_to_underscore($name))
+		or $self->emit_and_exit(1, "No Instant Answer found with name '$name'");
+	my $mod_name = $ia->{perl_module};
+	$mod_name =~ /^DDG::(?:Goodie|Spice)(?:::[^:]+)*::([^:]+)$/
```
zachthompson: I'm not sure why we're even checking the module.  It *must* be `$ia->{id}` atm.

GuiltyDolphin: Uhh... *shrug*

I think this is so we get the CamelCase name? It might not be required.

zachthompson: Well, the check and the error aren't quite right.  It's not checking whether it's in the root of a repo at all.  If we wanted to check the module name in metadata against that in the repo for consistency, that would be different.

GuiltyDolphin:
@zachthompson How's this? We allow access to a sub for looking up IAs by a name that can be an ID, name, package name etc. Rather than *only* allowing access to the name. We can then use this however we wish in the various commands.

GuiltyDolphin:
@zachthompson Anything else?

zachthompson:
Thank you.

moollaza:
@zachthompson @GuiltyDolphin I think this broke installs for me. It's now looking for `DDG::Publisher` to be installed on my local box

GuiltyDolphin:
@moollaza Uh, I'm good at breaking things :smile:

So the Travis build was fine after adding it back in?

moollaza:
> @moollaza Uh, I'm good at breaking things 😄

So I see... ;)

> So the Travis build was fine after adding it back in?

Looks like it! https://github.com/duckduckgo/p5-app-duckpan/pull/351

PR #318, opened 2016-03-15
Adds a hardcoded list of potential author ids to try.

kablamo:
# the problem

Sometimes we want to install old versions of modules.  duckpan doesn't handle that because 02packages.details.tar.gz only contains paths to the latest versions.  Paths are built from the id of the person who did the release and lots of different people do releases for DDG::Publisher.  

# a solution (kind of)

This is a terrible hack.  This patch helps duckpan cycle through a list of potential author ids. For each author id it builds a url and checks if the url points to a file which actually exists on duckpan.org.

Since the list of potential author ids is hardcoded, it will need to be updated manually when new authors come along.  So this doesn't totally solve the problem.

Note that this pr has an *additional* hack for Andrey's username which contains an _ which is not a valid character in PAUSE ids.  Which means CPAN::DistnameInfo can't parse the distribution name correctly.  A better solution would be to not allow underscores in duckpan ids, but I suspect thats hard to fix.

# testing

```
cpanm --uninstall DDG::Publisher
DDG_PUBLISHER=1110 duckpan DDG::Publisher
```

This test is particularly effective right now because alohaas released DDG-Publisher-1110 but has not released any other dists which causes duckpan to fail to find the module when the pin is 1110.

### DIFF
--- a/lib/App/DuckPAN/Perl.pm
+++ b/lib/App/DuckPAN/Perl.pm
@@ -191,16 +191,42 @@ sub duckpan_install {
 	}
 }
 
+my @POTENTIAL_AUTHORS = qw|
+    ALOHAAS
+    ANDREY_P
+    BRAD
+    BRIANS
+    CAINE
+    CRAZEDPSYC
+    DDGC
+    DYLANLL
+    GETTY
+    JAG
+    JBARRETT
+    JDORW
+    MGA
+    MOOLLAZA
+    RSSSSL
+    SDOUGBROWN
+    TOMMYTOMMYTOMMY
+    YEGG
+    ZT
+|;
+
 # We need to derive URLs because duckpan.org doesn't understand module@0.147 syntax
 sub find_previous_url {
 	my ($self, $module, $desired_version) = @_;
 
 	# Shaky premise #1: the author of our previous version is a current author.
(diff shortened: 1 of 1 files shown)


lib/App/DuckPAN/Perl.pm
```diff
@@ -191,16 +191,42 @@ sub duckpan_install {
...
+my @POTENTIAL_AUTHORS = qw|
+    ALOHAAS
+    ANDREY_P
+    BRAD
+    BRIANS
+    CAINE
+    CRAZEDPSYC
+    DDGC
+    DYLANLL
+    GETTY
```
moollaza: If there's nothing released by these people that we currently use, we can probably remove some of these names. Though it doesn't really hurt to keep them around?

- GETTY
- DYLANLL
- CRAZEDPSYC
- SDOUGBROWN

moollaza:
I think it would be safest to give Andrey a new username? "ANDREY" should be fine because we don't have any other "Andrey"'s.

> Note that this pr has an additional hack for Andrey's username which contains an _ which is not a valid character in PAUSE ids. Which means CPAN::DistnameInfo can't parse the distribution name correctly. A better solution would be to not allow underscores in duckpan ids, but I suspect thats hard to fix.

/cc @zachthompson for technical review

zachthompson:
Thanks @kablamo

PR #320, opened 2016-03-16
Updating Local Install instructions

moollaza:
Added steps to install Duckpan.org packages and fix any SSL issues.

Two users recently ran into issues and these were the steps I told them to follow which solved their problems.

/cc @zachthompson @talsraviv 

/cc @xaviervalarino

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,7 +281,7 @@ dzil authordeps | cpanm --mirror http://duckpan.org
 Additionally you may run into issues with SSL Verification. In this case installing Mozilla::CA should solve your problems:
 
 ```shell
-cpamn Mozilla::CA
+cpanm Mozilla::CA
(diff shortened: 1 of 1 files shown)


moollaza:
@zachthompson additionally I think we may want to cleanup and update the install.pl script to use `--mirror ...` and it looks like we no longer need to have users install `Crypt::SSLeay` according to https://metacpan.org/pod/Crypt::SSLeay#DO-YOU-NEED-Crypt::SSLeay

zachthompson:
Yeah, looks like you've started on install.pl.  Thank you.

talsraviv:
@moollaza this is for local dev environments right not codio?

moollaza:
Yup

PR #322, opened 2016-03-24
Add ID from metadata into structured_answer to emulate previous behaviour

moollaza:
It turns out we require an `id` to be defined in the JSON object we pass along to `DDH.add()`.

Now that we've removed the `id` from the Perl, this injects the ID from the metadata into the `structured_answer`.

### DIFF
--- a/lib/App/DuckPAN/Web.pm
+++ b/lib/App/DuckPAN/Web.pm
@@ -302,9 +302,13 @@ sub request {
 		foreach my $result (@results) {
 
 			my $caller = $result->caller;
+			my $id;
 			#exlude parent/child IAs like CheatSheets
 			if(my $ia = DDG::Meta::Data->get_ia(module => $caller)){
-				push @ids, $ia->[0]{id} if @$ia == 1;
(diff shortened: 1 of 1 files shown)


lib/App/DuckPAN/Web.pm
```diff
@@ -353,6 +357,7 @@ sub request {
 					my $structured = $result->structured_answer;
 					push @ids, $structured->{dynamic_id} if exists $structured->{dynamic_id};
 					if(exists $structured->{templates}){ # user-specified templates
+						$structured->{id} = $id if $id;
```
zachthompson: An id is required, no?

moollaza: @zachthompson I did that because `$id` wouldn't be defined above when Cheat Sheets are triggered, but it's OK because `$structured->{id}` is already defined for Cheat Sheets.

zachthompson: Got it! :+1:

zachthompson:
Thank you.

PR #323, opened 2016-03-26
DuckPAN new: reject existing instant answer names

### LINKED ISSUE #306: Duckpan new does not reject existing IAs
I think `duckpan new` should fail immediately when an existing instant answer is entered, rather than prompting the user to enter more details (and waste time).

GuiltyDolphin:
Reject an inputted name if it already belongs to an instant answer. Fixes #306.

Uses functionality defined in #317.


zachthompson:
Conflicts.

GuiltyDolphin:
@zachthompson That ended up being a very nice conflict resolution!

GuiltyDolphin:
@zachthompson I'll close this for now - it relies on removed functionality.

It should be fairly easy to implement with either #343 or #344 as we could just check if the IA already has files configured.

PR #329, opened 2016-04-12
Update DuckPAN to use DDG.ready intead of run-on-ready class

moollaza:
We ditched the `script-run-on-ready` class internally for `DDG.ready()`. This updated DuckPAN accordingly to use that

/cc @bsstoner @andrey-p @zachthompson


bsstoner:
:+1:

PR #330, opened 2016-04-12
Fix dists.ini Git::NextVersion for new 4 digit tags

moollaza:
This is required to create further release of DuckPAN with the correct release version


zachthompson:
Thank you.

PR #331, opened 2016-04-16
Add session caching to Web

GuiltyDolphin:
Certain assets such as `countries.json` and JavaScript dependencies were not being cached in `duckpan server`.

I've added caching to `App::DuckPAN::Web` to ensure that _within a single session_ (i.e, one load of `App::DuckPAN::Web`) the assets will not be requested more than once.

Load 1 (without caching): `17 requests, 1,666.72 KB, 26.09 s`.
Load 2 (without caching): `14 requests, 1,631.05 KB, 16.05 s`.
Load 3 (without caching): `14 requests, 1,631.05 KB, 16.37 s`.

Load 1 (with caching): `18 requests, 1,690.06 KB, 31.46 s`.
Load 2 (with caching): `14 requests, 1,631.05 KB 0.77 s`.
Load 3 (with caching): `14 requests, 1,631.05 KB, 0.81 s`.

My tests are a bit wacky, but in short: I had extreme delays before caching, with caching there is only a delay on the initial load. I have noticed no side-effects from caching.

/cc @jagtalon @moollaza @fzzr- (thought you might want to know).

### DIFF
--- a/lib/App/DuckPAN/Web.pm
+++ b/lib/App/DuckPAN/Web.pm
@@ -80,6 +80,7 @@ sub run_psgi {
 }
 
 my $has_common_js = 0;
+my %session_cache;
 sub request {
 	my ( $self, $request ) = @_;
 	my $hostname = $self->server_hostname;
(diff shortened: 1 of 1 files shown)


fzzr-:
Looks great for me. If I understand it correctly there will update cache only if request was fully success - simple and cool.

GuiltyDolphin:
@fzzr- Mhmm. We update the cache once per session (for each asset), on the first successful request.

zachthompson:
In general this looks like good idea.  Couple of things 

1. We should probably integrate this into the [caching functionality](https://github.com/duckduckgo/p5-app-duckpan/blob/master/lib/App/DuckPAN/Cmd/Server.pm) so the assets persist across restarts.
2. Rather than just downloading automatically on start, the assets should be mirrored.  Some of these will probably never change.

GuiltyDolphin:
@zachthompson :+1: Yeah - good ideas. I'll have a look into integrating with the existing caching; it'll probably work out better with a more persistent cache.

zachthompson:
Still working on this?

GuiltyDolphin:
@zachthompson Yeah - I've had a poke.

Run into a few issues in testing full-caching with my browser (keeps crashing); no issue on the current version though.

It seems a bit odd that we don't have a package/role specifically for caching (unless I've missed it); shouldn't we move all the caching functionality out into a new package?

zachthompson:
`Server` is really the only command that uses it.

GuiltyDolphin:
@zachthompson There are some caching definitions in `DuckPAN`, more in `Server` (a small amount in `Query`) and with this, `Web`.

moollaza:
Closing due to DDH entering [Maintenance Mode](https://duckduckhack.com/). We are now only accepting issues for essential bugs/fixes.

Issue #327, opened 2016-04-11
Incorrect detection of Instant Answer files


GuiltyDolphin:
Some commands, such as `duckpan server`, will pick up any `.pm` files in `lib/DDG/(Goodie/Spice/...)` and treat them as Instant Answer files; making it incompatible with multi-file Instant Answers (see https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1927#issuecomment-208041735). There is no issue if the Instant Answer is specified (see https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1927#issuecomment-208042511 and https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1927#issuecomment-208042609), as in `duckpan server MyIA`. 

/cc @mintsoft @zachthompson

Issue #200, opened 2015-01-08
Method to retrieve IA directories


zachthompson:
Would be handy to have a method (in App::DuckPAN?) that would return the directories involved with a particular IA.  These are in _build_ia_types in the respective templates key as far as I can tell.  But to be honest, it's not quite clear if these are the _only_ directories involved.

This would be used, for example, in finding out which directories need to be monitored for changes during development.

Issue #179, opened 2014-11-19
Duckpan command rm


killerfish:
@moollaza @mwmiller guys what do you think about having a command like `duckpan rm instant_answer_name`, i know its not going to be used that extensively, but in future, it might? Also since we have a new command, a rm command makes sense, no?

Issue #141, opened 2014-10-20
Duckpan Console


killerfish:
@moollaza @mwmiller I wanted to discuss this idea with you guys before diving into this completely. Basically i want to add a CLI to duckpan, but i'd like your input if this is really necessary or overkilling it. The "issues" this would address are:
- The transitioning between duckpan server/query and running tests, is a bit disjoint atm. Assuming a random sequence, I have run query then quit query, then run server, then quit server, run prove etc. I feel this could be improved. 
- Every time i have to make changes to a file, i have to switch dirs, then if i have to run/test it, i have to switch back to the root of repository. 
- Every time i make changes to a file (perl), i would have to quit query/server, and restart it. 
- We should be able to generate test files while testing under query.

For now these are the ones that come to mind, but these only involve duckpan query/server and prove. I have yet to explore the duckpan command(s) used for community platform though.

PR #335, opened 2016-04-30
Improve Templates

GuiltyDolphin:
### Goals

- [ ] Improve the general templating system.
- [x] Allow Cheat Sheets to be (mostly) configured through DuckPAN; this includes key/val, template type, sections etc.
- [X] Use IA pages to generate as much of the body as possible.
- [ ] Allow further configuration of existing Instant Answers.
- [X] Move templates into DuckPAN (we will need to update the repos); the templates are *directly* involved with DuckPAN, so they should probably be with the repo where we can update them en masse, and take advantage of this to better configure them to use better, more advanced features of Text::Xslate.

### Current Issues

N/A

### Bugs/issues fixed

- [X] Cheat Sheets getting a prompt for generic templates.
- [X] Cheat Sheet output file incorrectly named (currently cannot be fixed in DuckPAN, requires modification of Goodie template file). (Fixed in DuckPAN - but we may need to remove redundant template files in other repos).

- [ ] #141 - A 'console' interface for controlling aspects of DuckPAN.
- [ ] #179 - Add a command for removing Instant Answers (files).
- [ ] #200 - Allow retrieval of all directories/files for an Instant Answer.
- [ ] #327 - (Bug): Instant Answer files sometimes being retrieved incorrectly.


GuiltyDolphin:
@zachthompson Any thoughts on the general idea of this? I'm aiming to make it as *easy* as possible for a user to create as *much as possible* of the IA without having to do any hard work, then only have to modify the IA-specific functionality.

I think the use of the IA pages and metadata will allow us to generate most of the body.

GuiltyDolphin:
@zachthompson Hehe, I'm clearly trying to do too much here.

I'll work on pulling stuff out of this - there's several PRs in here...

GuiltyDolphin:
@zachthompson This is becoming a bit cumbersome, I'll close this for now and create issues where appropriate so we can better discuss the individual changes and decide what we want and don't want.

PR #338, opened 2016-05-04
Allow duckpan to be run from anywhere within an Instant Answer repository

### LINKED ISSUE #255: Error message about running DuckPAN from root of repo should be clarified
https://duck.co/forum/thread/13304/following-the-tutorial-for-creating-ia-having-erro

A user did not understand the error message we provide when someone attempts to run duckpan outside of an Instant Answer repo. We should re-word the message, and/or give an example `like zeroclickinfo-goodies/`

GuiltyDolphin:
Rather than failing whenever DuckPAN is run from within an Instant Answer repository, but not the root, we *find* the correct root then use that as the working directory for the remaining commands.

Fixes #255

/cc @zachthompson @moollaza

### DIFF
--- a/lib/App/DuckPAN.pm
+++ b/lib/App/DuckPAN.pm
@@ -156,10 +156,11 @@ sub _build_ia_types {
 	my $ddg_path = path('lib', 'DDG');
 	my $t_dir = path('t');
 	return [{
-			name      => 'Goodie',
-			dir       => $ddg_path->child('Goodie'),
-			supported => 1,
-			templates => {
+			name          => 'Goodie',
+			dir           => $ddg_path->child('Goodie'),
+			supported     => 1,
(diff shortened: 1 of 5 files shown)


zachthompson:
Cheat sheets not working with this for me.

GuiltyDolphin:
@zachthompson Could you elaborate? Where are they not working?

zachthompson:
@GuiltyDolphin weird, just tried again and it seems to be working.

GuiltyDolphin:
@zachthompson *shrug* Weird. Everything should basically work *exactly* the same as if DuckPAN was run in the root directory so if you notice any weird behavior let me know!

GuiltyDolphin:
@zachthompson I don't think there's anything else to be done - happy with this?

zachthompson:
Thank you.

Issue #222, opened 2015-02-10
More/Better debug output (perhaps a --debug flag?)


moollaza:
A new dev was trying to create a Spice IA that didn't need to return an answer to an API. He was unable to determine why his IA didn't appear in the output from `duckpan query` when he tested the triggers.

It turns out he had `return;` instead of `return "";` which obviously meant his IA was bailing out, so nothing was returned to the console.

I was thinking we should be able to indicate whenever an IA truly triggers, i.e. the `handle` is executed` but returns `undef` (meaning it bailed out)

This output might get really noisy though, so perhaps a special `--debug` flag is necessary?

@jagtalon @killerfish @MrChrisW @mintsoft any thoughts?

MrChrisW:
:+1: Yes please! I keep forgetting you can use `DDG_BLOCK_TRACE=1` flag to get more information :laughing: . A debug argument would make life easier.

mintsoft:
Absolutely; I always forget about `DDG_BLOCK_TRACE`; `--verbose=` or `--debug` is definitely more nix standard

killerfish:
i think we already have a --verbose, just add this under that flag perhaps?

PR #339, opened 2016-05-05
Add 'debug' flag as an alias for 'verbose'

GuiltyDolphin:
Fixes part of duckduckgo/p5-app-duckpan#222 as we can now use '--debug'
as a flag.

### DIFF
--- a/lib/App/DuckPAN.pm
+++ b/lib/App/DuckPAN.pm
@@ -80,7 +80,7 @@ option colors => (
 );
 
 option verbose => (
-	is      => 'ro',
+	is      => 'rw',
 	lazy    => 1,
 	short   => 'v|debug',
 	default => sub { 0 },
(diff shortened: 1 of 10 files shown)


GuiltyDolphin:
@zachthompson I'm finished with this, anything to add?

zachthompson:
Haven't had a chance to look at this really.  Looks like you changed a bit more than adding an alias ;)  Are you saying that all of these changed were necessary to support `--debug`?

GuiltyDolphin:
@zachthompson Heh, sorry, here's what I've done:

- Added a role for commands that support verbose. 

- Replaced standard warn/die with use of Carp.

GuiltyDolphin:
I've got some plans for Option::Tell in the future, but for now it allows switching of verbose/debug on any command.

zachthompson:
@GuiltyDolphin This is too big of a change for a simple alias.,

GuiltyDolphin:
@zachthompson Hehe, I'm not very good at naming these PRs :wink:

How about I pull the changes out into a couple of PRs? One for *just* the alias, one for the carp update, and either another for the per-command debug or have that with the first?

PR #340, opened 2016-05-12
Correct 'error' exit codes

GuiltyDolphin:
@zachthompson Instead of using '-1' (isn't that what C programmers do?) for exit
codes, use '1' to signal failure. From the docs itself: 'The only
Universally recognized values for {exit code} are 0 for success and 1
for error'.

### DIFF
--- a/lib/App/DuckPAN.pm
+++ b/lib/App/DuckPAN.pm
@@ -101,7 +101,7 @@ sub _build_duckpan_packages {
 	my $mirror_to   = $self->cfg->cache_path->child($gz);
 
 	if (is_error(mirror($package_url, $mirror_to))) {
-		$self->emit_and_exit(-1, "Cannot download $package_url");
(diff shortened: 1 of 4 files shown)


zachthompson:
Have to make sure we are checking codes correctly elsewhere.

GuiltyDolphin:
@zachthompson You mean in DuckPAN? All the codes are used for is the final exit in [emit_and_exit](https://github.com/duckduckgo/p5-app-duckpan/blob/master/lib/App/DuckPAN.pm#L367). I just checked with `exit -1`, and (at least on my term) that becomes the exit code `255`.

zachthompson:
@GuiltyDolphin No, I mean internally.  We use duckpan to install distributions in various places and I'll need to audit them to make sure they aren't checking for `-1` explicitly.

GuiltyDolphin:
@zachthompson M'kay!

zachthompson:
Thank you.

PR #341, opened 2016-05-14
Ensure Restart doesn't exit DuckPAN forcefully

GuiltyDolphin:
There are cases (like with Console - pending) where we don't want duckpan to exit
immediately, but rather we would like to return to a menu; now we may do
so.

Basically this ensures we can run Query as a sub-command.

/cc @zachthompson

### DIFF
--- a/lib/App/DuckPAN/Cmd/Query.pm
+++ b/lib/App/DuckPAN/Cmd/Query.pm
@@ -10,8 +10,7 @@ use MooX::Options protect_argv => 0;
 
 sub run {
 	my ($self, @args) = @_;
-
-	exit $self->run_restarter(\@args);
+	$self->run_restarter(\@args);
 }
 
 sub _run_app {
--- a/lib/App/DuckPAN/Restart.pm
+++ b/lib/App/DuckPAN/Restart.pm
@@ -24,7 +24,7 @@ sub run_restarter {
 	    defined(my $app = fork) or die 'Failed to fork application';
 	    unless($app){ # app kid
 	        $self->_run_app($args);
-	        exit 0;
+	        return;
 	    }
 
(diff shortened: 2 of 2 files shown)


zachthompson:
Query runs fine afaict.  What is the use case for this change?

GuiltyDolphin:
@zachthompson For any console interfaces to DuckPAN, because Restarter *explicitly* runs `exit`, it means that as soon as `Query` exits the whole script will too. Updating Restarter to just return means that we can run commands using Restarter within a larger (Console) interface without it forcing the application to exit.

zachthompson:
@GuiltyDolphin 
> run commands using Restarter within a larger (Console) interface

There is no use case I can think of where a user requests to quit query and we need to continue running, nor do I see this as a direction we need to consider atm.

GuiltyDolphin:
@zachthompson Mainly Re https://github.com/duckduckgo/p5-app-duckpan/issues/141, for example:

```

(w/ an IA loaded)

1 > Help
2 > Query
3 > Files
4 > Exit

> 2

...
Enter Query: 1 + 7
... 8
Enter Query: <Empty>
(bye!)

1 > Help
2 > Query
3 > Files
4 > Exit
```

But mhmm, okay - if this is a path we don't wish to take I'll close this

zachthompson:
@GuiltyDolphin Ah ok, that idea.  I see where you were going now.  We *may* want to do something like this in the future but in the near-term it will be low priority.

For future reference, you only need to worry about the last `exit` in Restart and the one in Query.  We want the child processes to exit.

GuiltyDolphin:
@zachthompson Okay! Let me know if we ever decide to implement Console - I've got a working version on one of my branches (though it depends on a bunch of other changes which we may not want).

PR #343, opened 2016-05-14
Add InstantAnswer Utilities

GuiltyDolphin:
Provides utilities for:

* Checking the type of Instant Answer (Spice, Goodie, Cheat Sheet etc.)

* Finds InstantAnswer files based on metadata (it'll get most things - but it'd be better if some of these were provided in the metadata directly).

* Provides a metadata lookup (see duckduckgo/duckduckgo#195 for an implementation that we would be able to use instead).

Depends on: #342 (or at least, supersedes it).

/cc @zachthompson

### DIFF
--- a/lib/App/DuckPAN/InstantAnswer/Util.pm
+++ b/lib/App/DuckPAN/InstantAnswer/Util.pm
@@ -99,11 +99,16 @@ sub find_ia_files {
 	} else {
 		# Back-end files
 		my $ddg_path = _path_from_package($ia->{perl_module});
+		my $back_end_base = $ia->{perl_module} =~ s/^DDG::[^:]+:://r;
+		my $back_end_separated = _path_from_package($back_end_base);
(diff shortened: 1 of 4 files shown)


lib/App/DuckPAN/InstantAnswer/Util.pm
```diff
@@ -0,0 +1,155 @@
...
+# ABSTRACT: Utilities for working with Instant Answer Metadata.
+
+BEGIN {
+	require Exporter;
+
+	our @ISA = qw(Exporter);
+
+	our %EXPORT_TAGS = (
+		predicate => [qw(is_cheat_sheet is_spice
+											is_goodie is_full_goodie)],
```
GuiltyDolphin: Well this is interesting indenting...

zachthompson:
@GuiltyDolphin Where is the pain point here?

PR #344, opened 2016-05-14
Add InstantAnswer::Config

GuiltyDolphin:
This provides a wrapper around the metadata from the platform; allowing us to provide additional information and operations that aren't present in the metadata.

Depends on #343.

/cc @zachthompson

### DIFF
--- a/lib/App/DuckPAN/InstantAnswer/Config.pm
+++ b/lib/App/DuckPAN/InstantAnswer/Config.pm
@@ -26,6 +26,11 @@ sub _build_files {
 	return \%files;
 }
 
+sub is_configured {
+	my $self = shift;
+	return @{$self->files->{all}} ? 1 : 0;
+}
+
 $Data::Dumper::Terse = 1;
 sub for_display {
 	my ($self, $item) = @_;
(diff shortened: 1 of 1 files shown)


zachthompson:
What operations have enough use to warrant looking into this?

GuiltyDolphin:
@zachthompson Yeah - sorry I haven't made this super clear.

I'm breaking up #335 into smaller pieces so we can pick and choose what we want.

A great use case for *this* in particular is when detecting files - instead of using templates to find which directories to monitor, we can grab each InstantAnswer::Config's files and monitor those. The main user of this is `Restarter`.

I'm still figuring out what goes with what, and the dependencies - so would you like me to outline all the changes in #335 and then you can say which features you think are acceptable?

If you think having Restarter use the files from InstantAnswer::Config is a good use case I can pull those changes into this PR.

GuiltyDolphin:
@zachthompson This would be very useful for commands like `duckpan rm`, `duckpan mv` etc. as we contain the 'meta' information along with files, making it easy to find which files need to change for the given command.

At the least I think we need to pull out the file-finding based on metadata (as in #343) so that we have a way of finding files independent of templates.

PR #345, opened 2016-05-18
Add 'debug' flag as an alias for 'verbose'

### LINKED ISSUE #222: More/Better debug output (perhaps a --debug flag?)

GuiltyDolphin:
Fixes #222 as we can now use '--debug' as a flag.

/cc @zachthompson

### DIFF
--- a/lib/App/DuckPAN.pm
+++ b/lib/App/DuckPAN.pm
@@ -82,7 +82,7 @@ option colors => (
 option verbose => (
 	is      => 'ro',
 	lazy    => 1,
-	short   => 'v',
+	short   => 'v|debug',
 	default => sub { 0 },
 	doc     => 'provide expanded output during operation',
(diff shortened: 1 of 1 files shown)


zachthompson:
Thank you.

PR #346, opened 2016-05-18
Use Carp for error messages

GuiltyDolphin:
Carp provides better error messages than just 'die' or 'warn'; this switches use of 'die' and 'warn' to 'croak' and 'carp' respectively.

/cc @zachthompson

### DIFF
--- a/lib/App/DuckPAN/Restart.pm
+++ b/lib/App/DuckPAN/Restart.pm
@@ -6,6 +6,7 @@ use Filesys::Notify::Simple;
 
 use App::DuckPAN::TemplateDefinitions;
 use Try::Tiny;
+use Carp;
 
 use strict;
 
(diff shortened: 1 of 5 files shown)


zachthompson:
`Carp` is mostly useful in the case where someone is **use**ing your module.  It tells them where in *their* code, rather than your module, the error happened (see discussion [here](http://www.perlmonks.org/?node_id=685452) for example.)

In this case, developers aren't doing that.  The errors and line numbers are for people like you and I to figure out where the problem is in DuckPAN. 

One borderline case is the auto-templating message in Web.  Yet it won't give what we want -  a reference to the line in their Goodie - but just a massive stack trace that is largely unhelpful for an IA dev.

GuiltyDolphin:
@zachthompson Ah, okay! My understanding was Carp was generally better/verbose output - that discussion clarifies things greatly!

There's a few cases in there (such as templates dying if files already exist) we may want to look at - it can't be super helpful for the users.

PR #350, opened 2016-06-01
Only check for IA repo when doing IA-specific commands

zachthompson:
Moving the check for an IA repo outside of `get_ia_type` broke commands like `release` which are used in non-IA repos.

cc @bsstoner @GuiltyDolphin

### DIFF
--- a/lib/App/DuckPAN.pm
+++ b/lib/App/DuckPAN.pm
@@ -632,7 +632,15 @@ sub checking_dukgo_user {
 
 sub get_ia_type {
 	my ($self) = @_;
-	return $self->repository;
+
+	my $repo = $self->repository;
+	unless ($repo) {
+		$self->emit_and_exit(-1,
+			'Must be run from within an Instant Answer repository'
+		);
+	}
+
+	return $repo;
 }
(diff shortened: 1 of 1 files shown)


GuiltyDolphin:
@zachthompson We should just be able to add `with 'App::DuckPAN::Option::Global';` to any commands that need to be runnable outside the repos.

zachthompson:
I prefer limiting the check to the one place it needs to happen rather than having to add an exclusion everywhere else. Was there a reason it needed to be in `initialize_working_directory`?

moollaza:
LGTM 👍

GuiltyDolphin:
@zachthompson Yeah, `initialize_working_directory` was to initialize the working directory for IA repos - so it shouldn't be encountered at all by non-IA commands (hence `Option::Global`); having the check in `initialize_working_directory` ensures IA commands aren't run outside of a valid repo directory.

The check makes sense if anything is ever going to try and access the repository (outside of `App::DuckPAN`); but that doesn't seem to be the case at the moment, so this *shouldn't* break anything.

PR #357, opened 2016-07-01
Check post_body is defined before modifying

moollaza:
It looks like on Codio (Perl 5.18) the line I've adjusted sees `$post_body` as uninitialized. I wasn't able to figure out why exactly that is. I couldn't determine any other cause so I assume its due to Perl 5.18?

Checking it's defined-ness seems to fix.

/cc @jbarrett @zachthompson 

codio_-_duckduckhack


jbarrett:
:+1:

PR #358, opened 2016-07-01
Add Fathead support for DuckPAN Query & Server

moollaza:
/cc @zachthompson @jbarrett @jdorweiler 

These changes add support to `duckpan query` and `duckpan server` for Fathead features used by open source Fatheads.

You can query the exact names of articles in the Output.txt file to see the result that will display. If an article is a redirect, it will follow the redirect and show the correct result. Disambiguations and Abstract Images are also supported.

Category Pages and External Links (infobox) support has not been implemented yet. But I plan to implement the latter soon. Also Fathead Triggers are not yet supported but I will add that soon as well, using the IA Page triggers. However, internally I don't believe we consider the triggers defined on the IA page (though we should!)

This approach ignores the DDG infrastructure and request handlers, which Spice and Goodies use. That system is heavily dependant on loading Fathead Perl modules into Blocks, but we removed the Perl modules long ago as they were only there to contain Fathead Metadata.

One other missing feature is the internal processing of the output.txt -- Internally we process each line of the output.txt and generate several variations of the Article title, e.g. for "background-color" we also create "background color" and "backgroundcolor".

I noticed that some Fathead parsing scripts implement this functionality themselves. We should probably make it known to developers that we'll do this work for them to simplify the parsers a bit.

Any feedback is welcome. I've not used DBI before so I wasn't sure about the best way to get the output.txt into a DB for searching. I've used **DBD::CSV** which interprets the output.txt (tab delimited) as a CSV and pulls it in to memory for SQL querying. I'm not sure how well this will work for very large output.txt files. We can easily test this though.

### How to Test

1. `git checkout` this branch
2. Clone the Fathead repository: `git clone git@github.com:duckduckgo/zeroclickinfo-fathead.git`
3. Start the server: `$ duckpan -I ../p5-app-duckpan/lib/ server <instant_answer_id>`
    e.g. `$ duckpan -I ../p5-app-duckpan/lib/ server mdn_css`
4. Enter one of the article titles from the associated Output.txt as a search query to see the content displayed.
  - e.g. "auth_basic" from https://github.com/duckduckgo/zeroclickinfo-fathead/blob/master/lib/fathead/nginx/output.txt

-------

To view the changes made while ignoring whitespace changes: https://github.com/duckduckgo/p5-app-duckpan/pull/358/files?w=0


zachthompson:
I'm not sure I understand the rationale of this being in duckpan.  You can pull out the keys, or just dump them in a debug statement in your fathead script, and you know *everything* it will trigger on exactly.

moollaza:
> I'm not sure I understand the rationale of this being in duckpan. You can pull out the keys, or just dump them in a debug statement in your fathead script, and you know everything it will trigger on exactly.

@zachthompson I'm more interested in getting it to work in DuckPAN Server so developers can actually see what the results will look like. I can look at the output.txt to get an idea of what content I'm showing, but being able to see the result is a big bonus and means you don't have to wait for us to deploy it to Beta to see it working. In fact, you don't even really need to write any code if you just manually create an output.txt and load it into DuckPAN for testing. If we add support for more Fathead templates this will become even more useful.

I agree that the duckPAN query integration doesn't offer as much benefit, but there's still something I enjoy about working with a REPL to see that the queries I'm thinking about trigger and return results. IMO looking at the data in aggregate can be a little overwhelming.

lib/App/DuckPAN/Fathead.pm
```diff
@@ -0,0 +1,57 @@
...
+with 'App::DuckPAN::HasApp';
+
+use DBI;
+use JSON;
+use Path::Tiny;
+use Data::Printer return_value => 'dump';
+
+sub search_output {
+
+    my ($self, $query, $output_txt) = @_;
```
GuiltyDolphin: @moollaza Tabs! :wink:

lib/App/DuckPAN/Query.pm
```diff
@@ -59,74 +61,96 @@ sub _default { warn "Unhandled event - $_[ARG0]" }
...
-	$k->yield('_get_user_input');
+    my ($k, $h, $query, $args) = @_[KERNEL, HEAP, ARG0, ARG2];
+
+    my ($app, $blocks) = @$h{qw{app blocks}};
+    Encode::_utf8_on($query);
+
+    my $repo = $app->get_ia_type;
+
+    try {
+        unless ($repo->{name} eq "Fathead") {
```
GuiltyDolphin: Hmm, I think if we used `if` here, then `else` it would be easier to change if we want to separate the behaviour of Spice and Goodies (or add other Instant Answers). Not a big issue though.

lib/App/DuckPAN/Web.pm
```diff
@@ -27,6 +28,7 @@ has page_js => ( is => 'ro', required => 1 );
 has page_locales => ( is => 'ro', required => 1 );
 has page_templates => ( is => 'ro', required => 1 );
 has server_hostname => ( is => 'ro', required => 0 );
+has repository => ( is => 'ro', required => 0 );
```
GuiltyDolphin: We don't use this anywhere.

lib/App/DuckPAN/Web.pm
```diff
@@ -276,6 +278,29 @@ sub request {
 		my @calls_goodie;
 		my @ids;
 
+		my $repo = $app->get_ia_type;
+
+		##########
+		# FATHEAD
```
GuiltyDolphin: Do we need this comment? The condition is pretty clear.

lib/App/DuckPAN/Web.pm
```diff
@@ -276,6 +278,29 @@ sub request {
...
+		if ($repo->{name} eq "Fathead") {
+
+			my $output_txt = path( $app->fathead_output() );
+			if ($output_txt->exists) {
```
GuiltyDolphin: Why not bail earlier?

If `fathead_output` and doesn't exist, we don't want to go through the effort of setting anything up. We should just kill the server immediately as there's nothing useful we can do.

We may wish to throw an informative error if they are in a non-FatHead repo and have specified `output_txt`.

(`"output_txt" is only valid for FatHead Instant Answers"`)

GuiltyDolphin: Why not bail earlier?

If `fathead_output` and doesn't exist, we don't want to go through the effort of setting anything up. We should just kill the server immediately as there's nothing useful we can do.

We may wish to throw an informative error if they are in a non-FatHead repo and have specified `output_txt`.

(`"output_txt" is only valid for FatHead Instant Answers"`)

GuiltyDolphin:
@zachthompson I agree with @moollaza on `duckpan server` - many developers may wish to observe the Instant Answer in an environment similar to what they would see on production; and Beta has a huge reliance on deploys. If we want to increase the number of developers working on FatHead Instant Answers I think this is a good move.

lib/App/DuckPAN/Fathead.pm
```diff
@@ -0,0 +1,57 @@
...
+
+    # Open output.txt file for searching
+    # Handles as a CSV with "\t" separator
+    # Provide numbered column names
+    my $dbh = DBI->connect ("dbi:CSV:", undef, undef, {
+        f_dir                   => $output_txt->parent,
```
GuiltyDolphin: This `HasApp`, so we can use `$self->app->fathead_output`.

lib/App/DuckPAN.pm
```diff
@@ -79,6 +79,13 @@ has duckpan_packages => (
 	builder => 1,
 );
 
+has fathead_output => (
```
GuiltyDolphin: We'd also want to ensure that if this is set, it gets normalized to a path.

lib/App/DuckPAN/Fathead.pm
```diff
@@ -0,0 +1,57 @@
...
+
+    eval {
+        my $sth = $dbh->prepare("SELECT * FROM output WHERE col1 = ?");
+        $sth->execute($query);
+        while (my $row = $sth->fetchrow_arrayref) {
+            $result = $row;
+        }
+        $sth->finish();
+    };
+    $@ and die "SQL database error: $@";
```
GuiltyDolphin: `$self->app->emit_and_exit(1, "SQL database error: $@")`

lib/App/DuckPAN/Fathead.pm
```diff
@@ -0,0 +1,57 @@
...
+    return $result;
+}
+
+sub output_as_json {
```
GuiltyDolphin: This a placeholder? Doesn't look like it is used anywhere (or does anything).

lib/App/DuckPAN/Query.pm
```diff
@@ -59,74 +61,96 @@ sub _default { warn "Unhandled event - $_[ARG0]" }
...
+                my $result = App::DuckPAN::Fathead->search_output($query, $output_txt);
+                if ($result){
+                    $app->emit_info('---', "Match found: $output_txt", p($result, colored => $app->colors), '---');
+                }
+                else {
+                    $app->emit_info('Sorry, no matches found in output.txt');
+                }
+            }
+            else {
+                $app->emit_info('Sorry, no output.txt file was not found');
```
GuiltyDolphin: We should probably not be continuing, is there anything useful we can do without `output_txt`?

lib/App/DuckPAN/Query.pm
```diff
@@ -5,52 +5,54 @@ use Moo;
...
-	$powh_readline->bind_key("C-\\", "interrupt");
-	$powh_readline->read_history($history_path);
-	$powh_readline->put('(Empty query for ending test)');
+    # Session that handles user input
+    my $powh_readline = POE::Wheel::ReadLine->new(
+        InputEvent => '_got_user_input'
+    );
+    $powh_readline->bind_key("C-\\", "interrupt");
+    $powh_readline->read_history($history_path);
+    $powh_readline->put('(Empty query for ending test)');
```
GuiltyDolphin: More spaces - tabs!

lib/App/DuckPAN.pm
moollaza: This gives me a way to access the `output_txt` option elsewhere (e.g Web.pm), without having to pass along new args from /cmd/server.pm and /cmd/query.pm.

AFAICT the `option` needs to be defined in `server.pm` and `query.pm` and not `DuckPAN.pm` otherwise the command would be:

`duckpan -o /path/to/output.txt server` which IMO feels awkward.

There may be a better solution, but this worked ;)

GuiltyDolphin: @moollaza Yes - unfortunately the command would have to be in that form.

May I suggest an alternative in the form of a Role then?

If you have (something like):

```perl
# lib/App/DuckPAN/Option/FatHead.pm
package App::DuckPAN::Option::FatHead;

use Moo::Role;
use MooX::Options;

with qw( App::DuckPAN::Cmd );

option output_txt => (
    ...
    trigger => \&_initialize_fathead_output,
);

sub _initialize_fathead_output {
    my $self = shift;
    $self->app->fathead_output(path($self->output_txt));   
}
```

Then you could just do `with qw( App::DuckPAN::Option::FatHead )` whenever you need to use it.

lib/App/DuckPAN.pm
```diff
@@ -275,6 +275,17 @@ sub _build_ddg {
 	App::DuckPAN::DDG->new( app => shift );
 }
 
+has fathead => (
```
GuiltyDolphin: What's this for? Why does DuckPAN have a FatHead?

moollaza: This allows DuckPAN to load the `App::DuckPAN::Fathead` into app.

Combined with `HasAPP` it enables access to the Fathead lib where needed, via `$self->app->fathead->`.

I was contemplating renaming the lib to something like `App::DuckPAN::Util::Fathead` or similar. We load Perl.pm and DDG.pm in the same fashion for the same reasons.

moollaza: Or I could just rename this to `fathead_util`... Same result

GuiltyDolphin: @moollaza Ah, so it is utility functions for FatHeads rather than information on the particular FatHead being run on?

jbarrett:
Is this ready to start experimenting with? How can we help you test it?

I think there are a bunch of proposed fatheads in the pipeline we can start using as test load :)

GuiltyDolphin:
@moollaza Could we use some of the information from the IA metadata (and ID) to generate some of the paths?

For example, rather than having to specify `duckpan server -o lib/fathead/perl6_doc/output.txt` you would just say `duckpan server Perl6Doc` or `duckpan server perl6_doc` and it would figure the rest out. Admitedly FatHeads are a bit of an edge-case with a different directory structure to Goodies and Spice, but this should be easily do-able.

moollaza:
> Is this ready to start experimenting with?

@jbarrett Yes! I've been using for the past week or so. At a very basic level it's working, but without the connection to Metadata it's not as good as it could be. 

> How can we help you test it?

Try testing any open Fathead PR's with this to see what works and what doesn't. I don't love specifying the output.txt path and would like to improve this

> Could we use some of the information from the IA metadata (and ID) to generate some of the paths?

@GuiltyDolphin that's what I'd like to do, and your suggestion is exactly what I had in mind. It just feels weird specifying the name of a Perl Module which doesn't actually exist. IMO we need a better way to connect Fathead files in the repo to their metadata, because we're relying on the name of an imaginary Perl module which doesn't exist in the repo or anywhere else.

We should probably change "Perl Module" on the IA Page to "Directory" or similar which maps to a dir in https://github.com/duckduckgo/zeroclickinfo-fathead/tree/master/lib/fathead -- that's something real that the developer has knowledge of

@zachthompson @jbarrett @jdorweiler @GuiltyDolphin what are your thoughts on that?

moollaza:
I just realized we assume the `/lib/fathead/` dirs are named according to the Fathead IDs. That should make things easier 👍

moollaza:
I've now added support for Metadata and remove the `-o` option.

Now you just specify the Fathead ID as the last argument:

e.g. `duckpan server mdn_css`

We're operating on the assumption that a dir exists in `/lib/fathead/` with the same name which means we can easily check for the existence of an `output.txt` file in the dir

lib/App/DuckPAN/DDG.pm
```diff
@@ -14,39 +14,49 @@ sub show_failed_modules {
...
-	    @args = map {
-				my $camel_name = $self->app->get_ia_by_name($_)->{perl_module};
-			} @args;
+		my @fatheads = grep {
+			$type->{name} eq "Fathead"
+		} @args;
+		$self->app->fathead->selected(@fatheads);
+
+		my @clean_args = grep {
+			$type->{name} eq "Spice" || $type->{name} eq "Spice"
```
GuiltyDolphin: This appears to be the exact same check, twice?

lib/App/DuckPAN/Fathead.pm
```diff
@@ -0,0 +1,205 @@
...
+
+}
+
+# Attempt to get a result from DB (output.txt)
+# Capture & display any raised errors
+sub _db_lookup {
+	my ($self, $query) = @_;
+
+	my $result;
+	$@ = "";
```
GuiltyDolphin: `$@ = ''`

GuiltyDolphin:
@moollaza This appears to be triggering on everything now? If I try `foo` with the `perl_doc` FatHead it returns a result with all the meta information but no text (as there is no field for `foo`).

moollaza:
> This appears to be triggering on everything now?

Yup, there was a bug. Should be all fixed now 👍

moollaza:
I think this is pretty good for now. It does not yet Category Pages, but AFAICT no open source Instant Answers use Category pages.

I haven't added support for External Links (infobox) yet either, but will once he mdn_css Fathead starts using them. I don't think any other IAs currently use the external links.

moollaza:
Added a few more commits and updated the PR description

GuiltyDolphin:
@moollaza Getting a `Can't locate object method "_get_image" via package "App::DuckPAN::FatHead" .../lib/App/DuckPAN/FatHead.pm line 173.` in `duckpan query`.

moollaza:
@GuiltyDolphin whoops, didn't mean to change that function's name yet. Fixed

jagtalon:
Looks like the URL is not showing in the "More at ..." even thought it can read it in DuckPAN:

screen shot 2016-07-27 at 3 03 07 pm
screen shot 2016-07-27 at 3 03 16 pm

jbarrett:
For some reason I was working off the impression this had been released and opened an issue: #365 - Once you've gotten a match, that match will always display until the next match.

jagtalon:
I thought it had been released too :)) It took me an embarrassingly long time to figure that out.

lib/App/DuckPAN/Fathead.pm
```diff
@@ -0,0 +1,257 @@
...
+	predicate => 1,
+	trigger => 1
+);
+
+sub _trigger_selected {
+	my ( $self, $id ) = @_;
+	my $dir = path("lib/fathead/$id");
+	unless ($dir->is_dir) {
+		my $full_path = $dir->realpath;
+		$self->app->emit_and_exit(1, "Directory not found: $full_path") ;
```
jbarrett: Should this check for the presence of output.txt within the directory? It will bomb out if you haven't yet run your parser.

GuiltyDolphin:
@moollaza No support for related topics? Any chance we could get support for this soonish?

moollaza:
> No support for related topics? Any chance we could get support for this soonish?

@GuiltyDolphin I don't believe any existing IAs use it so I didn't have anything to test with. I was going to add support once the in-progress MDN CSS Fathead started using them ;)

lib/App/DuckPAN/Fathead.pm
moollaza: @jbarrett I actually check below in the `output_txt` builder if the file exists. I suppose I could do that check in here and define `output_txt` here as well, removing the need for the builder?

Looking back, I don't recall why I used a builder for `output_txt`

moollaza:
> I thought it had been released too :)) It took me an embarrassingly long time to figure that out.

If everyone's already using this, then it sounds like there aren't any major issues with it?

@zachthompson care to merge? Happy to keep refining this as necessary.

lib/App/DuckPAN/Fathead.pm
jbarrett: Since it's lazy loaded, it doesn't check for output.txt until it's first accessed (I think), so the server will run and then bail when you search.

GuiltyDolphin:
@moollaza Okay, well it's not a killer atm, but we're getting a fair few FatHeads now which may wish to use related topics, so it should definitely be one of (if not the) next features to be implemented.

Otherwise, this looks good to merge - I've been using it a fair bit and it seems to work pretty well.

lib/App/DuckPAN/Fathead.pm
```diff
@@ -105,10 +105,10 @@ sub _search_output {
 
 	my $result = $self->_db_lookup($query);
 
-	if ($result && $result->{type} eq "R") {
+	while ($result && $result->{type} eq 'R') {
```
moollaza: @GuiltyDolphin I don't expect this will ever run more than once, but thanks for fixing the bug here!

moollaza:
Merging with @zachthompson's approval 👍

lib/App/DuckPAN/Fathead.pm
GuiltyDolphin: @moollaza I've run into a couple of situations where it wants to run more than once; but yeah, probably not very common.

lib/App/DuckPAN/Fathead.pm
```diff
@@ -216,7 +216,7 @@ sub _parse_disambiguations {
 			my $text = $root->as_trimmed_text;
 
 			# Build URL Path
-			my $href = "/".$out->{meta}->{src_id}."/$title";
+			my $href = "/?q=$title&ia=about";
```
moollaza: 👍

jagtalon:
Woo!

PR #359, opened 2016-07-02
Add template var for cheat sheet file names

### LINKED ISSUE #353: Underscores in new cheat sheets
When creating a skeleton cheat sheet with `--cheatsheet` it separates terms with an underscore in the name of the JSON file.  That's almost always wrong except in the rare cases when the underscore is part of the "thing" the cheat sheet is about. 

cc @GuiltyDolphin

GuiltyDolphin:
@zachthompson This adds a template variable that can be used for the cheat sheet file name so we have hyphens rather than underscores.

Fixes #353.

This needs to be merged *before* duckduckgo/zeroclickinfo-goodies#3337.

### DIFF
--- a/lib/App/DuckPAN/Cmd/New.pm
+++ b/lib/App/DuckPAN/Cmd/New.pm
@@ -212,6 +212,12 @@ sub run {
 		ia_path_lc        => $lc_filepath,
 	);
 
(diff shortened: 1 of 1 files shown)


zachthompson:
Thank you.

PR #361, opened 2016-07-11
Pass Spice call_data to JS Spice object

moollaza:
## Background

There was some recent discussion asking if we could formalize the weird script tag hack we have in place for grabbing extra data passed along from the Perl. (i.e. we parse the api call `<script>` tag to pull out useful information, like the parsed query)

I don't want us to formalize that weird hack and remembered that I had experimented with the `call_data` parameter before. Now that we're passing Perl data to the front end for Goodies I decided to take another stab at this, using the [`DDG.inject`](https://github.com/duckduckgo/duckduckgo-utils/blob/master/util.js#L325) function, which we use internally in a similar fashion.
## Result

This should allow us to easily pass extra data from Perl along to a front end Spice IA object.

I've tested this and it seems to be working correctly, getting it to work internally will really prove my theory.
## Testing

To test this, I modified the NPM Spice's `handle` function to:

``` perl
handle remainder_lc => sub {
    return unless $_;
    return $_, data(
        is_this_working => "yes!",
        query => $req->query
    );
};
```

And updated the JS to log the `call_data` with `console.log(Spice.npm.call_data);` which worked as expected:

npm_colors_at_duckduckgo_and_1____projects_zeroclickinfo-spice__perl__and_any_do

The code for the modified NPM Spice is on GitHub here: https://github.com/duckduckgo/zeroclickinfo-spice/compare/zaahir/spice-call-data-example

You can test this PR by checking out that Spice branch and running `duckpan -I ../p5-app-duckpan/lib server`, the triggering the NPM Spice with `npm colors`. The console will contain the Spice Data object passed from the Perl

---

@zachthompson I'd like to hear your thoughts on this

### DIFF
--- a/lib/App/DuckPAN/Web.pm
+++ b/lib/App/DuckPAN/Web.pm
@@ -273,6 +273,7 @@ sub request {
 		my @calls_nrc = ();
 		my @calls_script = ();
 		my %calls_template = ();
+		my %calls_data = ();
 		my @calls_goodie;
 		my @ids;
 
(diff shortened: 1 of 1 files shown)


moollaza:
@zachthompson this is now ready for review

moollaza:
Closing. We don't have plans to use call_data in production.

PR #367, opened 2016-08-02
Remove output_txt builder function, move into _trigger_selected

moollaza:
Minor cleanup here to simplify checking for `output.txt` and containing dir

@jbarrett is this more in line with your expectation?

### DIFF
--- a/lib/App/DuckPAN/Fathead.pm
+++ b/lib/App/DuckPAN/Fathead.pm
@@ -26,28 +26,20 @@ sub _trigger_selected {
 		my $full_path = $dir->realpath;
 		$self->app->emit_and_exit(1, "Directory not found: $full_path") ;
 	}
+	my $file = $dir->child("output.txt");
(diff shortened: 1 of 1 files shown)


lib/App/DuckPAN/Fathead.pm
```diff
@@ -26,29 +26,21 @@ sub _trigger_selected {
...
+	unless ($file->exists){
+		my $full_path = $file->realpath;
+		$self->app->emit_and_exit(1, "No output.txt was not found in $full_path");
+	}
+	$self->output_txt($file);
 	return $dir;
 }
 
 has output_txt => (
 	is => 'rw',
```
GuiltyDolphin: Why don't we make this `rwp`? Then we can use `$self->_set_output_txt(...)`.

I doubt anything else would need to change the `output_txt`?

jbarrett:
:+1: Thanks!

Sorry for delay!

PR #368, opened 2016-08-03
FatHeads - add support for trigger words

GuiltyDolphin:
This allows better emulation of the triggering on production by providing a means of specifying trigger words.

Currently just accepts a `trigger_words.txt` file in the same directory as `output.txt` (not required) for trigger words; we could potentially have a CL option as well.

cc @moollaza @jbarrett

### DIFF
--- a/lib/App/DuckPAN/Fathead.pm
+++ b/lib/App/DuckPAN/Fathead.pm
@@ -64,8 +64,8 @@ has _trigger_re => (
 sub _build__trigger_re {
 	my ($self) = @_;
 	my @words = @{$self->_trigger_words};
-	my $text = '\b(?:' . (join '|', map { quotemeta $_ } @words) . ')\b';
-	return qr/$text/i;
+	my $text = join '|', map { quotemeta $_ } @words;
+	return qr/\b(?:$text)\b/i;
 }
 
 has output_txt => (
--- a/lib/App/DuckPAN/Fathead.pm
+++ b/lib/App/DuckPAN/Fathead.pm
@@ -37,21 +37,19 @@ sub _trigger_selected {
 
 has _trigger_words => (
 	is      => 'ro',
-	lazy    => 1,
 	builder => 1,
+	lazy    => 1,
 );
 
 sub _build__trigger_words {
 	my ($self) = @_;
 	my $tf = 'trigger_words.txt';
-	my $file;
-	if ($self->has_selected) {
-		$file = path("lib/fathead/", $self->selected, $tf);
-		unless ($file->exists){
-			my $full_path = $file->realpath;
-			$self->app->emit_error("No $tf was found in $full_path");
-			return [];
-		}
+	return [] unless $self->has_selected;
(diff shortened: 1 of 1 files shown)


jbarrett:
Nice! I think trigger_words.txt is a convention we just made up during perldoc development, so a CL option would be cool for other ones... Or we could just adopt trigger_words.txt as a thing.

GuiltyDolphin:
@jbarrett I thought it would be a pretty good convention to start (and the file doesn't *need* to be comitted for it to work), as @jdorweiler said:

> @jbarrett How about put the trigger words in a separate file trigger_words.txt? I'll have to add them manually to our internal list of triggers but at least they will be documented in a place where everyone can see them.

It would help document *how* you expect the FatHead to trigger.

But sure, I can add an option for triggers.

jdorweiler:
You're right, I just made `trigger_words.txt` up :smile: 
I'm going to start asking everyone to include one though.

moollaza:
I was actually going to add support for trigger words via the Meta data. We already have a place to define them on the IA Pages so I figured we might as well use that?

@jdorweiler thoughts?

jdorweiler:
Maybe @zachthompson remembers but I tried adding trigger words to the triggers field on the IA page and it broke some stuff. I don't know if that was ever fixed.

lib/App/DuckPAN/Fathead.pm
```diff
@@ -29,6 +29,39 @@ sub _trigger_selected {
...
+	is      => 'ro',
+	lazy    => 1,
+	builder => 1,
+);
+
+sub _build__trigger_re {
+	my ($self) = @_;
+	my @words = @{$self->_trigger_words};
+	my $text = '\b(?:' . (join '|', map { quotemeta $_ } @words) . ')\b';
+	return qr/$text/i;
```
moollaza: nit: I think it would be clearer to put the `\b` and `(?:` into the `qr/`

```perl
my $text = join '|', map { quotemeta $_ } @words;
return qr/\b(?:$text)\b/;
```

moollaza:
> I tried adding trigger words to the triggers field on the IA page and it broke some stuff.

@jdorweiler you mean simply have the IA page trigger words filled out created problems? I have them filled out for https://duck.co/ia/view/mdn_css, and it seems fine? That's what I was going to use to get it working via metadata.

lib/App/DuckPAN/Fathead.pm
GuiltyDolphin: Sure.

GuiltyDolphin:
@moollaza Hmm, metadata would be a good option - @jdorweiler it shouldn't be an issue if we ensure triggers are reviewed before deploy? At the least we could use it in 'testing'.

GuiltyDolphin:
@jdorweiler @moollaza @jbarrett What do you think would be the best way to do this?

- Provide a command-line options for trigger words
- Support a trigger-word file
- Use the metadata (currently cannot access triggers though).

moollaza:
@GuiltyDolphin Thanks! This LGTM :+1:

PR #371, opened 2016-08-06
Add support for Fathead related topics

moollaza:
Added support to properly parse related topics list and generate data for InfoBox.

Our internal JS takes the `RelatedTopics` array and creates the final `data` array required by the Infobox template.

Note: It looks like the new Ruby IA is the only open source IA currently using Related Topics

/cc @jdorweiler @zachthompson @GuiltyDolphin

### DIFF
--- a/lib/App/DuckPAN/Fathead.pm
+++ b/lib/App/DuckPAN/Fathead.pm
@@ -189,7 +189,10 @@ sub _build_structured_answer {
 		);
 	}
 
+
 	# Category Pages Result
+	# TODO:
+	#  Build category pages by parsing each article's categories
 	if ($data->{type} eq 'C') {
 		$out->{duckbar_topic} = 'List';
 		$out->{model} = 'FatheadListItem';
--- a/lib/App/DuckPAN/Fathead.pm
+++ b/lib/App/DuckPAN/Fathead.pm
@@ -169,9 +169,12 @@ sub _build_structured_answer {
 		%extra_data = (
 			Heading 	=> $data->{title},
 			Abstract 	=> $self->_replace_newlines($data->{abstract}),
-			AbstractURL => $data->{abstract_url},
-			FirstURL 	=> $metadata->{src_url},
+			AbstractURL	=> $data->{abstract_url},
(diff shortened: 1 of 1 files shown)


moollaza:
Result:

array_try_convert_at_duckduckgo

GuiltyDolphin:
@moollaza Looking at the 'internal' processing, it seems quite contrary to what the docs state:

> Related topics. These will be turned into links in the Zero-click Info box. Examples: [[Perl Data Language]]. If the link name is different, [[Perl Data Language|PDL]].

But you've used the `external_links` to generate the related topics? And using a `\n` syntax rather than brackets?

I'm confused - what *exactly* is used to build the related topics? Do the docs need clarifying (I've not been able to get them to work with the info from the docs - only chance and... Oddities)?

lib/App/DuckPAN/Fathead.pm
```diff
@@ -231,6 +237,26 @@ sub _parse_disambiguations {
...
 }
 
+# Parse list of Related Topics into hash
+# Processed by internal JS to create InfoBox w/ "Related Topics" heading
+sub _parse_related_topics {
+	my ($self, $related_topics, $out) = @_;
+	my @infobox_data;
+	my @related_topics = split /\\n/, $related_topics;
+	foreach my $related_topic (@related_topics){
+		if ($related_topic =~ m/^\[\[(.+)(?:|(.+))?\]\]$/) {
```
GuiltyDolphin: Uhh, how about `Regexp::Common` for balanced brackets? Would make sense internally too.

moollaza:
> I'm even more confused now... I might've been on an out-of-date branch?

@GuiltyDolphin I incorrectly implemented external links as related topics. I realized after how external links work (they're all passed through the API, but only the first is used as a link along side the More at).

Related Topics are used to build an Infobox with links.

GuiltyDolphin:
@moollaza Ah, okay - fair dos.

Doesn't look like the related topics are working on Beta? Only categories seem to get displayed in `Related Topics`.

PR #372, opened 2016-08-10
Strip closing <script> tags from queries

moollaza:
We've made similar internal changes as a means of preventing XSS for now.

This enable us to no longer require Goodie output to be encoded with `html_enc`.

The production changes are already live. Now we just need to cleanup all the Goodies 😄 

/cc @mintsoft @GuiltyDolphin

### DIFF
--- a/lib/App/DuckPAN/Web.pm
+++ b/lib/App/DuckPAN/Web.pm
@@ -478,19 +478,21 @@ sub request {
 		my $calls_script = join('', map { q|<script type='text/JavaScript' src='| . $_ . q|'></script>| } @calls_script);
 		# For now we only allow a single goodie. If that changes, we will do the
 		# same join/map as with spices.
-		if(@calls_goodie){
+		if(@calls_goodie) {
 			my $goodie = shift @calls_goodie;
+			my $json = encode_json($goodie);
 			$calls_nrj .= "DDG.duckbar.future_signal_tab({signal:'high',from:'$goodie->{id}'});",
-			# Uncomment following line and remove "setTimeout" line when javascript race condition is addressed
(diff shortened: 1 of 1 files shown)


lib/App/DuckPAN/Web.pm
```diff
@@ -539,6 +541,19 @@ sub request {
 	return $response;
 }
 
+
+sub run_on_ready_script {
+	my ($str) = @_;
+
+	# any unescaped script tags at this stage will XSS the page - so strip them
+	$str =~ s|</?script>||gi;
```
GuiltyDolphin: @moollaza There could be other stuff in the `script` tag - we should use something like:

```perl
my $dom = Mojo::DOM->new->parse($str);
$dom->find('script')->map('strip');
$str = $dom->to_string;
```

moollaza: Content inside should be fine, we really just need to remove any `</script>` tags so that the `<script>` tag the JSON is inside of doesn't get closed prematurely.

mintsoft: I guess `<script/>` won't cause any problems as browsers won't expand it out? It should nest even if it did, but maybe worth a test?

zachthompson:
Is this just for parity with prod?  Who but the dev running duckpan would execute one of these?

lib/App/DuckPAN/Web.pm
GuiltyDolphin: @moollaza I get very nervous w/ Regex/HTML; it's just asking for trouble...

Something like `Mojo` should catch all the weird edge-cases w/o us having to worry any more; I stand by my statement.

moollaza:
> Is this just for parity with prod? Who but the dev running duckpan would execute one of these?

@zachthompson yes, for parity with Prod. Without this, you'd see breakage inside DuckPAN for queries containing `</script>`

PR #373, opened 2016-08-22
Add instructions for mac users if they failed to installl Crypt::SSLeay

soleo:
Add docs for instructing mac os users to find their OpenSSL library when installing duckpan

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,6 +261,34 @@ Additionally you may run into issues with SSL Verification. In this case install
 cpanm Mozilla::CA
 ```
 
+#### Crypt::SSLeay (for Mac OS El Capitan)
+
+If you failed to install Crypt::SSLeay with cpanm
+
+```shell
+cpanm Crypt::SSLeay
+--> Working on Crypt::SSLeay
+Fetching http://www.cpan.org/authors/id/N/NA/NANIS/Crypt-SSLeay-0.72.tar.gz ... OK
+Configuring Crypt-SSLeay-0.72 ... N/A
+! Configure failed for Crypt-SSLeay-0.72. See /Users/soleo/.cpanm/work/1471743850.28096/build.log for details.
+```
+and the build.log shows missing openssl
+```shell
+openssl-version.c:2:10: fatal error: 'openssl/opensslv.h' file not found
+#include <openssl/opensslv.h>
+         ^
+1 error generated.
+Failed to build and link a simple executable using OpenSSL
+-> N/A
+```
+
+Try install openssl with homebrew and set up library path
+
+```shell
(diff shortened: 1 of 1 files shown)


moollaza:
I was actually in the process of removing the `Crypt::SSLeay` dependency as it should no longer be necessary:

https://github.com/duckduckgo/p5-duckpan-installer/pull/5

moollaza:
Are you interested in taking over that PR?

soleo:
@moollaza funny you asked, I actually read https://github.com/duckduckgo/p5-duckpan-installer/pull/5/ before making this PR. I did fork the p5-duckpan-installer because I was hoping to simply the dockerfile. A couple of questions here. 

1) Any reason that we must run under non-root user?
2) https://github.com/duckduckgo/p5-duckpan-installer/pull/5/ is almost half year old now, are there anything which is no longer applied to current duckPAN anymore?
3) Is it safe to do https://github.com/soleo/p5-duckpan-installer/commit/e91767137fb5e3f729cfd7831d91b2c4298f1fe6 ? I added the mirror URL and notest flag to cpanm while installing.

moollaza:
> 1) Any reason that we must run under non-root user?

I'm not sure why it restricts the root user -- this script was written several years ago and hasn't changed in the past 4 or 5.

@zachthompson do you have any objections to removing the non-root user requirement here?

> 2) duckduckgo/p5-duckpan-installer#5/ is almost half year old now, are there anything which is no longer applied to current duckPAN anymore?

I suspect a lot of the dependencies we're adding here may no longer be used. Any Perl deps that are required should be added to DuckPAN's dist.ini so that installing via CPAN works the same as this script.

For example it appears Starman is used for `duckpan publisher` so we should probably add that dependency to the dist.ini since `AutoReqs` won't identify that as a requirement.

There's lots of room for improvement in this script, it just hasn't been a big issue and local DuckPAN installs are less popular now that we offer Codio.

> 3) Is it safe to do soleo/p5-duckpan-installer@e917671 ? I added the mirror URL and notest flag to cpanm while installing.

I think we'd want to know up front about any errors when installing? If it's suppressed and the dev runs into issues, it'll be harder to know later what's going on

soleo:
@moollaza probably not a good idea to run as root after I read http://stackoverflow.com/questions/4938592/how-why-does-npm-recommend-not-running-as-root/4999441.

zachthompson:
Right, it's never a good idea to do development as root.

soleo:
@zachthompson if you're running dev environment inside of a docker container, it's actually not an extremely bad idea. After all, anything happens inside of the container stays in container.  😄

PR #374, opened 2016-08-23
Ensure metadata is checked when no Instant Answers are specified

GuiltyDolphin:
We were only performing a metadata existence check when IAs were specified (thus `duckpan query` did no checks); this PR fixes that.

cc @moollaza @zachthompson


edgesince84:
thanks Ben

zachthompson:
Thank you.

PR #376, opened 2016-09-19
Rename LICENSE.md to LICENSE

moollaza:
Adjusting this as the license file is plain text, not markdown


moollaza:
I see now that having a `LICENSE` files creates an issues as `D::Z::Plugin::License` tries to create this file based on the `license` in the `dist.ini`

We can probably remove that from the `dist.ini`?

PR #377, opened 2016-09-21
remove Before::Build step, DDG installed via cpan.org

moollaza:
We've now released the DDG repo to CPAN, so the DDG::Meta::Dependency can be installed as needed when DuckPAN installed. 

My attempted BeforeBuild fix can now be removed.

I've updated the list of DDG::* dependencies to be skipped as well.

/cc @zachthompson @jbarrett

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -1,5 +1,5 @@
 name    = App-DuckPAN
-author  = Torsten Raudssus <torsten@raudss.us> L<https://raudss.us/>
+author  = DuckDuckGo <open@duckduckgo.com>
 license = Apache_2_0
 copyright_holder = DuckDuckGo, Inc. L<https://duckduckgo.com/>
 copyright_year   = 2013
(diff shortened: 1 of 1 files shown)


dist.ini
```diff
@@ -36,10 +36,7 @@ tag_format = %v
 push_to = origin master
 
 [AutoPrereqs]
-skip = ^DDG::*
-
-[Run::BeforeBuild]
-run = cpanm DDG --skip-installed --mirror http://www.cpan.org/ --mirror http://duckpan.org
+skip = ^DDG::(Spice|Goodie|Fathead|Longtail|Publisher)
```
GuiltyDolphin: Might we want to add this for roles too? `^DDG::((Spice|Goodie|FatHead|LongTail)(Role)?|Publisher)`.

There also appears to be a `DDG::Util` hierarchy in Spice.

moollaza: DuckPAN doesn't use/require any Role's so they don't need to be ignored as dependencies.

Technically Fathead and Longtail don't need to be here, but it doesn't really hurt.

Spice, Goodie and Publisher are all `use`d though, so it tries to install them

GuiltyDolphin: @moollaza Ah, alright! :+1:

PR #381, opened 2016-10-16
Failing to load metadata should be a warning

mintsoft:
At the moment if an IA can't be found then the IA isn't loaded. This switches it out so it will still attempt to load the IA if the metadata is unavailable.

FWIW @moollaza thought this was already the case

### DIFF
--- a/lib/App/DuckPAN/DDG.pm
+++ b/lib/App/DuckPAN/DDG.pm
@@ -77,8 +77,7 @@ sub get_blocks_from_current_dir {
 		else {
 			$failed_to_load{$arg} =
 				'Could not retrieve metadata - please ensure the Instant Answer page ' .
-				'is in development status or later.';
-			next;
+				'is in development status or later; the IA has still been loaded';
 		}
 		my ($load_success, $load_error_message) = try_load_class($class);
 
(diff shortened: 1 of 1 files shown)


lib/App/DuckPAN/DDG.pm
```diff
@@ -77,8 +77,7 @@ sub get_blocks_from_current_dir {
 		else {
 			$failed_to_load{$arg} =
 				'Could not retrieve metadata - please ensure the Instant Answer page ' .
-				'is in development status or later.';
-			next;
+				'is in development status or later; the IA has still been loaded.';
```
zachthompson: How about?

1. Use `emit_notice` to warn that it's not in metadata
2. Set `$class = $arg` which will fail if using an ID.

mintsoft: Will do

moollaza: Working on this in https://github.com/duckduckgo/p5-app-duckpan/pull/383

mintsoft:
@moollaza closing this in favour of https://github.com/duckduckgo/p5-app-duckpan/pull/383

PR #382, opened 2016-11-18
Add Fathead support to DuckPAN Test

moollaza:
- Allow `duckpan test` to check if named Instant Answer ID has a valid `/lib/fathead` dir containing an `output.txt`
- If so, modify `$ENV` accordingly for `validate_fathead.t`

/cc @jbarrett

### DIFF
--- a/lib/App/DuckPAN/Cmd/Test.pm
+++ b/lib/App/DuckPAN/Cmd/Test.pm
@@ -1,6 +1,7 @@
 package App::DuckPAN::Cmd::Test;
 # ABSTRACT: Command for running the tests of this library
 
+no warnings 'uninitialized';
 use MooX;
 with qw( App::DuckPAN::Cmd );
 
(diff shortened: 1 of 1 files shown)


lib/App/DuckPAN/Cmd/Test.pm
```diff
@@ -1,6 +1,7 @@
 package App::DuckPAN::Cmd::Test;
 # ABSTRACT: Command for running the tests of this library
 
+no warnings 'uninitialized';
```
zachthompson: Has to be below the moo junk or in `run()`.

moollaza: Thanks 👍 

I haven't been able to produce those warnings -- how did you do it?

zachthompson: Just test something non-existent.

zachthompson:
Thank you.

PR #383, opened 2016-11-21
Remove Metadata requirement for IAs to run in DuckPAN

moollaza:
## Background
Devs shouldn't **_need_** an IA page in dev status to play with DuckPAN. 

With the switch to the Programming Mission, devs are currently force to open PRs to test any new IAs, which means they are required to open PRs for non-mission IAs as well, which bloats the repo with PRs.

We should enable DuckPAN to use temporary metadata in order to trigger local IA's and show results.

This will help keep the repos clean with only PRs for IAs that are real and intended to move forward.

## Changes
- Update DDG.pm to gracefully handle IAs without metadata and use a temporary `ID` based on the Perl module
- Update Web.pm to generate "fake" metadata when none exists to be passed along to frontend JS to create IA tab and display result
  - Also, notify user when no metadata was found, and that temporary metadata is being used instead. Provide better explanation of how to get real metadata and IA Page working
- Minor fix to Server.pm that fixes warning

/cc @mintsoft

### DIFF
--- a/lib/App/DuckPAN/Cmd/Server.pm
+++ b/lib/App/DuckPAN/Cmd/Server.pm
@@ -190,7 +190,7 @@ sub change_js {
 }
 
 # Rewrite all relative asset links in CSS
-# Capture leading quote, insert
+# Capture leading quote, insert $hostname, append filename
 # E.g url("/assets/background.png") => url("http://duckduckgo.com/assets")
 sub change_css {
 	my ( $self, $css ) = @_;
--- a/lib/App/DuckPAN/Cmd/Server.pm
+++ b/lib/App/DuckPAN/Cmd/Server.pm
@@ -191,7 +191,7 @@ sub change_js {
 
 # Rewrite all relative asset links in CSS
 # Capture leading quote, insert $hostname, append filename
-# E.g url("/assets/background.png") => url("http://duckduckgo.com/assets")
+# E.g url("/assets/background.png") => url("http://duckduckgo.com/assets/background.png")
 sub change_css {
 	my ( $self, $css ) = @_;
 	my $hostname = $self->hostname;
--- a/lib/App/DuckPAN/Cmd/Server.pm
+++ b/lib/App/DuckPAN/Cmd/Server.pm
@@ -195,8 +195,7 @@ sub change_js {
 sub change_css {
 	my ( $self, $css ) = @_;
 	my $hostname = $self->hostname;
-
-	$css =~ s|:\s*url\(["']([^'"]+)["']\)|:url\("http://$hostname/$1")|gi;
+	$css =~ s!:\s*url\((["'])?/!:url\($1http://$hostname/!g;
 	return $css;
 }
 
--- a/lib/App/DuckPAN/Cmd/Server.pm
+++ b/lib/App/DuckPAN/Cmd/Server.pm
@@ -190,11 +190,13 @@ sub change_js {
 }
 
 # Rewrite all relative asset links in CSS
+# Capture leading quote, insert
 # E.g url("/assets/background.png") => url("http://duckduckgo.com/assets")
 sub change_css {
 	my ( $self, $css ) = @_;
 	my $hostname = $self->hostname;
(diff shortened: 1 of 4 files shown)


moollaza:
@zachthompson @mintsoft I've pushed some commits. This now prevents the message and DuckPAN Query works fine.

However, for DuckPAN Server, we have a few more direct calls to `DDG::Meta::Data` to grab the JS for Spices and the `id` for Goodies.

Any ideas on how we can bypass these? 

These are the lines that we need to adjust:
- https://github.com/duckduckgo/p5-app-duckpan/blob/master/lib/App/DuckPAN/Web.pm#L342
- https://github.com/duckduckgo/p5-app-duckpan/blob/master/lib/App/DuckPAN/Web.pm#L477

zachthompson:
@moollaza Think we'd need to add module-to-id mapping back to duckpan.  When the first lookup fails, derive the id from the module, push it, and it *should* work for the `get_js()` call.

mintsoft:
@moollaza someone new on slack ran into this today

moollaza:
@zachthompson this is now ready for further review.

/cc @mintsoft @GuiltyDolphin 

I have tested this by creating a new Goodie and Spice without metadata and running `duckpan server`, and `duckpan query`. In both cases the correct results were displayed.

Here's a screenshot of a "request headers" Spice that has no Metadata, showing a result:

2__zeroclickinfo-spice__duckpan_-e_-i____p5-app-duckpan_lib_server__perl__and_request_headers_at_duckduckgo

lib/App/DuckPAN/Cmd/Server.pm
```diff
@@ -190,11 +190,13 @@ sub change_js {
...
 # Rewrite all relative asset links in CSS
-# E.g url("/assets/background.png") => url("http://duckduckgo.com/assets")
+# Capture leading quote, insert $hostname, append filename
+# E.g url("/assets/background.png") => url("http://duckduckgo.com/assets/background.png")
 sub change_css {
 	my ( $self, $css ) = @_;
 	my $hostname = $self->hostname;
-	$css =~ s!:\s*url\((["'])?/!:url\($1http://$hostname/!g;
+
+	$css =~ s|:\s*url\(["']([^'"]+)["']\)|:url\("http://$hostname/$1")|gi;
```
moollaza: This fixes a separate bug I ran into...

zachthompson: Which is what?

lib/App/DuckPAN/DDG.pm
```diff
@@ -74,11 +92,21 @@ sub get_blocks_from_current_dir {
 		if (my $ia = $self->app->get_ia_by_name($arg)) {
 			$class = $ia->{perl_module};
 		}
+		elsif ($arg =~ /_/) {
```
zachthompson: What about IDs with no underscore?

lib/App/DuckPAN/Cmd/Server.pm
moollaza: The problem was related to our parsing and re-writing of relative urls in the CSS files we pull down from duckduckgo.com.

After looking into this more, I've realized that re-writing the CSS isn't even necessary because Web.pm proxies relative urls to the appropriate upstream host.

Removing the body of this function has no effect on the SERP AFAICT.

I will remove these changes

lib/App/DuckPAN/DDG.pm
moollaza: Good catch 👍 

I will check if it matches `/^[a-z0-9_]+$/` which resembles a lowercased ID.

If there are uppercase letters we can assume it's a Perl Module Name.

moollaza:
@zachthompson I've revised the logic here I think I've now accounted for any kind of valid input and it now bails on invalid inputs:

1____projects_zeroclickinfo-goodies__zsh_

moollaza:
Here's an example of all the cases:


1____projects_zeroclickinfo-goodies__zsh_

zachthompson:
Thanks.

PR #385, opened 2016-12-09
Create HTTP::Headers obj from headers hash

moollaza:
## Background
We've added support for Spices to use multiple headers, instead of just the `accept_header`.

The changes to DuckDuckGo are awaiting merge here: https://github.com/duckduckgo/duckduckgo/pull/208

The only remaining Spice, using the `accept_header` property has been updated here: https://github.com/duckduckgo/zeroclickinfo-spice/pull/3056

**Note:** This requires a new Perl module: `HTTP::Headers`

## Description
This PR updates DuckPAN to support passing along the `$headers` hash reference when creating the `HTTP::Request` object, by instantiating a `HTTP::Headers` object

## Testing

This can be tested by installing the newest DDG build, and running the DuckPAN Server in the Spice repo. The `Skyscanner` and `CocoaPods` Spices have already been updated to use the new `spice headers` property.

/cc @jbarrett

### DIFF
--- a/lib/App/DuckPAN/Web.pm
+++ b/lib/App/DuckPAN/Web.pm
@@ -15,6 +15,7 @@ use HTML::TreeBuilder;
 use HTML::Element;
 use Data::Printer return_value => 'dump';
 use HTTP::Request;
+use HTTP::Headers;
 use LWP::UserAgent;
 use URI::Escape;
 use JSON;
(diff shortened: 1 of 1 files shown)


moollaza:
We've now merged + released the headers changes in the DDG lib. We can move forward with testing + releasing these changes 👍

PR #387, opened 2017-03-22
Pass spice headers to GET requests and print HTTP::Request in DuckPAN output

moollaza:
When dealing with and debugging Spice IA's that use POST request and/or `spice headers` the developer is not able to see the POST request body, or the headers that are sent.

This minor change ensures that the full request for both GET and POST requests is displayed to the developer.

Also, this ensure that `spice headers` are also passed along when used with a GET request.

/cc @pjhampton 

---

Example output:

1__zeroclickinfo-spice__duckpan_-i____p5-app-duckpan_lib_server__gcat_


zachthompson:
ty

PR #388, opened 2017-05-05
DuckPAN Enhancement: Add FE tests to Testing tool.

pjhampton:
I have enhanced the `duckpan test <args>` command to include the front-end tests (if they exist) for a certain IA or all IAs. These run before the perl tests.

/cc @zachthompson

### DIFF
--- a/lib/App/DuckPAN/Cmd/Test.pm
+++ b/lib/App/DuckPAN/Cmd/Test.pm
@@ -80,28 +80,27 @@ sub run {
 		};
 
 		# For the front-end tests
-		print("\n\nSearching for front-end tests\n");
+		$self->app->emit_info("Searching for front-end tests");
 		if(scalar(@args) == 0) {
+
 			$self->app->emit_error("Couldn't run front-end tests. Please ensure that jasmine is installed globally.\n\nnpm install -g jasmine-node\n") if system("jasmine-node --test-dir spec/");
 		}
-		else {
+		elsif($ia_type eq "Goodie" or $ia_type eq "Spice") {
+
 			foreach my $ia_name (@args) {
+				my $test_file = "spec/$ia_name\_spec.js";
 
-				if($ia_type eq "Goodie" or $ia_type eq "Spice") {
-					my $test_file = "spec/$ia_name\_spec.js";
-
-					if(-e $test_file) {
-						print "\nRunning front-end tests for $ia_name\n";
-						$self->app->emit_error("Couldn't run front-end tests. Please ensure that jasmine is installed globally.\n\nnpm install -g jasmine-node\n") if system("jasmine-node $test_file");
-					}
-					else {
-						print "\nNo front-end tests detected for $ia_name\n";
-					}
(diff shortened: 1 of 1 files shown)


lib/App/DuckPAN/Cmd/Test.pm
```diff
@@ -70,14 +70,38 @@ sub run {
...
+				else {
 					$self->app->emit_and_exit(1, "Could not find output.txt for $id in $path");
 				}
 			}
 
 			$self->app->emit_and_exit(1, "Could not find any tests for $id $ia_type") unless @to_test;
 		};
 
+		# For the front-end tests
+		print("\n\nSearching for front-end tests\n");
```
moollaza: We use `$self->app->emit_info` for convenience messages 👍

lib/App/DuckPAN/Cmd/Test.pm
```diff
@@ -70,14 +70,38 @@ sub run {
...
+		if(scalar(@args) == 0) {
+			$self->app->emit_error("Couldn't run front-end tests. Please ensure that jasmine is installed globally.\n\nnpm install -g jasmine-node\n") if system("jasmine-node --test-dir spec/");
```
moollaza: Let's put this error text in an array so it's easier to read/edit and de-duplicates.

See https://github.com/duckduckgo/p5-app-duckpan/blob/390d5779d46e1ca154b38ca2e308a1ea04ae1216/lib/App/DuckPAN/DDG.pm#L103

lib/App/DuckPAN/Cmd/Test.pm
```diff
@@ -70,14 +70,38 @@ sub run {
...
+		}
+		else {
+			foreach my $ia_name (@args) {
+
+				if($ia_type eq "Goodie" or $ia_type eq "Spice") {
```
moollaza: IA type is set based on the repo duckpan is running inside. We can do this check once. Replace the `else` with and `elsif` for this condition -- I think that'll work...

lib/App/DuckPAN/Cmd/Test.pm
moollaza: Also I think we want to use \` instead of `system` so we see the output?

mintsoft: FYI: You'll have to use `$?` for the exitcode rather than the value of the `` operator though in that case

pjhampton: thanks guys 👍

moollaza:
@pjhampton this looks pretty good. I think it needs to prompt the user to run frontend tests, or perhaps only run them when it sees they are there.

In the spice repo I see a bunch of errors because there aren't any frontend tests:

1____projects_zeroclickinfo-spice__cat_

Might be ideal to add `--fe` and `--no-fe` to also force the tests to run or be skipped?

pjhampton:
Closing due to lack of attention

PR #389, opened 2017-07-07
Corrected regex for CamelCase package names

moollaza:
User reported that a new IA, `DDG::Spice::AnimeTest` wasn't loading.

I've updated the package name regex to properly handle CamelCase names so that uppercase letters can follow lowercased letters 👍 

/cc @pjhampton (this may have affected you recently as well!)


pjhampton:
nice @moollaza!

